PR #1, opened 2017-06-08
Travis CI

Br3nda:


.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+rvm:
+  - 2.3.3
+before_script:
+  - RAILS_ENV=test bundle exec rake db:create db:schema:load
+  - gem install brakeman bundler-audit
+  - bundle-audit update
+script:
+  - bundle exec rubocop -D
+  - brakeman --run-all-checks --exit-on-warn --format plain .
+  - bundle exec rspec spec
```
ben-rabid: You are not calling `bundle-audit` here, just the update above

ben-rabid: Not sure if you need the ` spec` at the end :elephant:

PR #3, opened 2017-06-08
change browsers supported

KatieWright26:
Flash warning appears based on what browsers are not supported - tweaked browsers to match what is in our support agreement


Br3nda:
CI will pass once #1 is merged

config/initializers/devise.rb
```diff
@@ -7,6 +7,7 @@
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
   # config.secret_key = '....'
+  config.secret_key = "ENV[DEVISE_SECRET_KEY]"
```
Br3nda: It should get this from config.secret_key_base. This key isn't set on production so it will fail there.

Br3nda: Try setting in your secrets.yml - or setting in your environment.

We should probably set up figaro gem for this in the future.

PR #7, opened 2017-06-09
Release 1.0

Br3nda:
Releasing our staging work to production


KatieWright26:
Is it problematic that the secret_key_base is displayed here?

Br3nda:
the key displayed is only for example environments (dev)
but maybe we should make the user generate one?

PR #15, opened 2017-06-12
Release 2

Br3nda:
This PR sends all our Staging work into production


Br3nda:
oops! this was the wrong way around

PR #23, opened 2017-06-22
Bump rubocop up to match hound

Br3nda:


Br3nda:
Noteworthy change here.
There's new cops now, that we aren't matching yet. Hound will tell us about these as we go. So i've turned off rubocop check in travis.

lib/tasks/sign_image_tasks.rake
```diff
@@ -1,7 +1,7 @@
 require 'fileutils'
 
 namespace :sign_images do
-  dimensions_regex = Regexp.new(/(\d{1,4})x(\d{1,4})/)
+  DIMENSIONS_REGEXP = Regexp.new(/(\d{1,4})x(\d{1,4})/)
```
eoinkelly: :elephant: FYI you don't need the `Regexp.new` to create a regex literal - the forward ticks creates a regex literal by itself

Br3nda: so, just `/(\d{1,4})x(\d{1,4})/`
?

Br3nda: i'm 99% sure this rake task doesn't run anywhere anymore

eoinkelly: yep

Br3nda:
"49 fixed issues". says codeclimate

PR #25, opened 2017-06-22
Small haml clean up

Br3nda:
Mostly wrapping line. If we don't, we'll get hound complaining later


Br3nda:
Maybe these can be in helpers to make.iy easier. They're in several places in the code.

PR #28, opened 2017-06-23
Release 4

Br3nda:
Summary of this release:

* Cleans up (removes) a bunch of files and mentions of flowplayer. This is a video player we used to use but don't anymore.
* Removes dependency on the AWS SDK gem. this was added when we moved some videos to S3 but it isn't needed to just serve files from S3. We were getting almost daily upgrade notifications for this when we don't even need it - so removed.

This release is currently on staging. Needs a test and then we can release to production.
http://nzsl-staging.herokuapp.com


Br3nda:
That merge conflict is odd!

PR #49, opened 2017-07-14
Release 6

Br3nda:


Gemfile
```diff
@@ -15,8 +14,7 @@ gem 'whenever'
 
 gem 'browser'
 gem 'ckeditor_rails'
-gem 'formtastic', '~>2.3.0'
-gem 'foundation-rails'
+gem 'foundation-rails', '~> 6.3.1.0'
```
elspeth-rabid: 6.3.1 is a bit buggy - I find 6.4.1.2 is the most reliable version available at the moment but NZSL doesn't use any of the Foundation fancy things so this is 100% 🐘

config/initializers/assets.rb
```diff
@@ -10,12 +10,12 @@
 # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
 # Rails.application.config.assets.precompile += %w( search.js )
 
-Rails.application.config.assets.precompile += %w( print.css )
-Rails.application.config.assets.precompile += %w( admin.css )
-Rails.application.config.assets.precompile += %w( application.css )
-Rails.application.config.assets.precompile += %w( print_screen.css )
-Rails.application.config.assets.precompile += %w( smoothness/jquery-ui.css )
-Rails.application.config.assets.precompile += %w( modernizr.js )
+Rails.application.config.assets.precompile += %w(print.css)
```
elspeth-rabid: Rails.application.config.assets.precompile += %w(admin.css print.css application.css)

Tider? 🐘

config/initializers/devise.rb
```diff
@@ -1,6 +1,7 @@
 # Use this hook to configure devise mailer, warden hooks and so forth.
 # Many of these configuration options can be set straight in your model.
 Devise.setup do |config|
+  config.secret_key = 'dabd54db3b77f2b880cffc1e90c85270063db8c636eb3280dd0c81bf40f633c3901a5c3d0ec04cb5965c2ab9aceac8e39ea08ad6b7dd3f431d2b4651080e586b' # rubocop:disable Metrics/LineLength
```
elspeth-rabid: This value is overwritten at the end of this config file in the production ENV.
Perhaps it would be clearer if the secret key settings were together? 🐘

config/locales/simple_form.en.yml
```diff
@@ -0,0 +1,31 @@
+en:
```
elspeth-rabid: I doubt any of these locale settings are used anymore.

config/initializers/devise.rb
Br3nda: :+1: good catch. Let's record that as an issue to work on.

PR #50, opened 2017-07-14
Fixing up code climate test reporting

Br3nda:


Br3nda:
Hoping this will fix up the coverage reporting for #48

PR #56, opened 2017-08-09
Add Google Tag Manager

elspeth-rabid:
Show tags only on production.
Remove Google Analytics because it is done from GTM now.
Hotjar integration to be managed there.


calebkeene:
🔥 🔥 🔥

PR #57, opened 2017-08-10
Feature/testing banner

elspeth-rabid:
image

I know it is wrong to have this at the top of the page but that is what Micky asked for.


vinithanatarajan:
screen shot 2017-08-14 at 11 12 47 am

The links are not highlighted blue in colour.

app/views/shared/_test_banner.html.haml
```diff
@@ -0,0 +1,10 @@
+.banner
+  .banner__content
+    %p
+      Hello and welcome! We have been working hard to improve the NZSL dictionary website.
+      We would love your feedback. As you go through the website, you will see questions pop up.
+      At the end of your visit, please click on
+      %a{ href: "https://insights.hotjar.com/s?siteId=587430&surveyId=26313" } this link
+      to give us more feedback. Thank you!
+    %p
+      %a{ href: "https://youtu.be/D2d-_OVhdog" } View this message in NZSL
```
Br3nda: should this be a `link_to` instead?
:elephant:

Br3nda: also -- should it be a new window or some other warning that you're leaving this website..?

Br3nda:
It's good to slowly chip away at the old style ruby hashes -- but there's so many and already so much hound things you've fixed so :elephant:

app/views/shared/_test_banner.html.haml
elspeth-rabid: @Br3nda - good thinking, I'll make it a target=_blank

PR #59, opened 2017-08-14
Make the images and links align with the bottom of the equalised sear…

elspeth-rabid:
image

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -18,4 +18,4 @@
 //= require foundation-equalizer
 //= require video
 
-$(document).foundation();
+$(function(){ $(document).foundation(); });
--- a/app/views/signs/search.html.haml
+++ b/app/views/signs/search.html.haml
@@ -31,9 +31,13 @@
               = link_to "learn more", sign_url(sign.id), {class: "show_link float-left"}
               %span.unicode_arrow.float-left &rarr;
             .drawing_container.float-left
-              = link_to(sign_url(sign.id), 
+              = link_to(sign_url(sign.id),
                         class: 'drawing grid_drawing js-ga-link-submission',
-                        onclick: "_gaq.push(['_trackEvent', 'Search', 'Click', '#{query_text}-#{sign.id}', #{@signs.find_index(sign)  + 1}]);") do
+                        onclick: "_gaq.push(['_trackEvent',
+                          'Search',
+                          'Click',
+                          '#{query_text}-#{sign.id}',
+                          #{@signs.find_index(sign) + 1}]);") do
                 = image_tag "signs/150-100/#{sign.drawing}"
 
   .pagination_container.small-12.float-left.text-center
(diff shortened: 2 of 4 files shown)


app/views/signs/search.html.haml
```diff
@@ -26,13 +26,13 @@
...
-                    %span.unicode_arrow.float-left &rarr;
-                  .drawing_container.float-left
-                    = link_to sign_url(sign.id), {:class => 'drawing grid_drawing js-ga-link-submission', onclick: "_gaq.push(['_trackEvent', 'Search', 'Click', '#{query_text}-#{sign.id}', #{@signs.find_index(sign)  + 1}]);"} do
-                      = image_tag "signs/150-100/#{sign.drawing}"
+          .link_and_drawing_container.small-12
+            .link_container.float-left
+              = link_to "learn more", sign_url(sign.id), {class: "show_link float-left"}
+              %span.unicode_arrow.float-left &rarr;
+            .drawing_container.float-left
+              = link_to sign_url(sign.id), {:class => 'drawing grid_drawing js-ga-link-submission', onclick: "_gaq.push(['_trackEvent', 'Search', 'Click', '#{query_text}-#{sign.id}', #{@signs.find_index(sign)  + 1}]);"} do
```
Br3nda: Can we use the new ruby hash syntax?

Br3nda: I don't think it needs to be a hash at all, it can just be
` +              = link_to sign_url(sign.id), class:  'drawing grid_drawing js-ga-link-submission', onclick: "_gaq.push(['_trackEvent', 'Search', 'Click', '#{query_text}-#{sign.id}', #{@signs.find_index(sign)  + 1}]);" do `

some of the complexity of this line can be simplified in a helper.
:elephant:  because it was already a mess before.

app/assets/javascripts/application.js
```diff
@@ -18,4 +18,4 @@
 //= require foundation-equalizer
 //= require video
 
-$(document).foundation();
+$(function(){ $(document).foundation(); });
```
Br3nda: my javascript is pretty rusty -- so i can't tell why does that need changing?

app/views/signs/search.html.haml
elspeth-rabid: Ha yeah I only changed the indentation, I didn't edit the actual code, but I can change it yip.

Br3nda (approved):

app/views/signs/search.html.haml
```diff
@@ -31,9 +31,13 @@
...
               %span.unicode_arrow.float-left &rarr;
             .drawing_container.float-left
-              = link_to(sign_url(sign.id), 
+              = link_to(sign_url(sign.id),
                         class: 'drawing grid_drawing js-ga-link-submission',
-                        onclick: "_gaq.push(['_trackEvent', 'Search', 'Click', '#{query_text}-#{sign.id}', #{@signs.find_index(sign)  + 1}]);") do
+                        onclick: "_gaq.push(['_trackEvent',
+                          'Search',
+                          'Click',
+                          '#{query_text}-#{sign.id}',
```
Br3nda: should this be in "" instead? Is it support to be string interpretation?

Br3nda: oh, nvm -- it is -- but the wrapping makes it hard to see

PR #61, opened 2017-08-29
Handled the query when value is nil.

vinithanatarajan:
This PR fixes this Raygun error https://app.raygun.com/crashreporting/w693o4/errors/2064502032#19748248658


Br3nda:
@vinithanatarajan don't forget to press "delete branch" button after merging.

PR #62, opened 2017-08-30
fix #148385435

vinithanatarajan:
Recuced the z-index of the play button on the weetbix cards so it doesn't overlap with the dropdown menu.

### DIFF
--- a/app/assets/stylesheets/_buttons.scss
+++ b/app/assets/stylesheets/_buttons.scss
@@ -58,7 +58,7 @@
   @include center;
   color: white;
   font-size: 70px;
-  z-index: 100;
+  z-index: 50;
 }
 
 a .button,


Br3nda:
What's #148385435 ?

vinithanatarajan:
#148385435 is the id for the bug in pivotal tracker

calebkeene:
@vinithanatarajan as an aside, your branch name is quite unusual - I would expect something like `fix/thing-being-fixed` or `bugfix/thing-being-fixed` or maybe even `bug/thing-being-fixed`. Call me pedantic, but even branch names are important :man_shrugging:

calebkeene:
@vinithanatarajan oh I just read it properly - I see that the word `fix` is in your description, so makes sense. Did you still want a review on this? I can give it a tick if you're waiting for one

vinithanatarajan:
@calebkeene ok I will keep the branch naming convention in mind. Yes, please review it.

PR #67, opened 2017-08-31
Promote staging code to production

Br3nda:


vinithanatarajan:
@Br3nda hound bot comment is fixed in this PR https://github.com/rabid/nzsl-online/pull/68
please take a look.

Br3nda:
Looks like Houndbot caught some bad haml  -- that might cause breakage. I don't think we have good tests of the admin UI - so this is good that hound found it.

app/views/layouts/admin.html.haml
```diff
@@ -31,9 +31,10 @@
       #admin-dashboard-sidebar.admin-dashboard-sidebar.position-left.off-canvas.off-canvas-absolute.reveal-for-medium.data-off-canvas
         .admin-dashboard-sidebar-title-area
           .admin-dashboard-open-sidebar
-            %button#open-sidebar.admin-dashboard-open-sidebar-button.show-for-medium{'aria-label': "open menu", type: "button"}
-              %span{'aria-hidden': "true"}
-                %a{href: "#"}
+            %button.admin-dashboard-open-sidebar-button
+            .show-for-medium{ 'aria-label': "open menu", type: "button" }#open-sidebar
```
Br3nda: i don't think this wrapped line is the same thing at all.
it was a button with a class of "show-for-medium" -- and it's changed to a button with nothing in it and a div after it.

Br3nda: image

It'll need fixing - it'll break admin if we release this

Br3nda: Proposed a fix in #69

PR #72, opened 2017-09-08
Fix up old style ruby hash in haml files

Br3nda:
I ran a subl plugin over the haml files, to fix up the old style ruby hashes


Br3nda:
I think all the hounds are now about lines being too long

jamesotron:
Hound has a lot to say about these files.

Br3nda:
@jamesotron yes, it's all "line too long" which we can fix next -- this PR fixed up all the old hash styles only. and it's getting big now.

PR #73, opened 2017-09-12
Task remove tracking requests

keyaral:
Tracking all the requests is causing the database to fall over. 
Removing the tracking code until the client wishes to do more analytic work.


Br3nda:
this will solve an immediate problem -- but it was added for a good reason: to know where the slowness was coming from.
Is it poss to instead recod the time taken to a log file?

There's also the admin interface -- which will be abandoned ghost code if we merge this PR.

Br3nda:
another approach, that lets us keep this audit trail, it to have a heroku scheduler task that deletes db rows > 7 days old. Heroku schedulers are free IIRC

elspeth-rabid:
Hey @Br3nda ! Were you thinking of making search request tracking controllable from the admin dashboard (in the CMS) or with .env/applications secrets?
I just implemented the latter and then thought 🤔

elspeth-rabid:
@Br3nda - I am using an environment variable for enabling/disabling request tracking and a rake task to your with Heroku Scheduler to keep the number of rows to a reasonable amount. The site is about to go down again - what do you think of my fix?

app/models/sign.rb
```diff
@@ -65,19 +65,18 @@ def self.current_page(per_page, last_result_index, all_result_length)
...
   def self.search(params)
-    before_time = Time.now.to_f
+    before_time = Time.current
     url = url_for_search(params)
     xml_document = Nokogiri::XML(open(url))
     entries = xml_document.css(ELEMENT_NAME)
     count = xml_document.css('totalhits').inner_text.to_i
-    save_time_elapsed(url, before_time, count)
-
+    save_time_elapsed(url, before_time, count) if Rails.application.secrets.track_search_requests?
```
joshmcarthur: Can this go in as a guard clause on `save_time_elapsed`?

lib/tasks/control_search_requests.rake
```diff
@@ -0,0 +1,19 @@
...
+
+  desc 'Limits the requests in the database to the most recent and limit the total number'
+  task manage_requests: :environment do
+    total_destroyed = 0
+    # Delete requests over STORAGE_DAYS days old
+    total_destroyed += Request.where('created_at < ?', STORAGE_DAYS.days.ago).destroy_all.length
+
+    # Ensure we have no more than MAX_REQUESTS of the most recent requests
+    if Request.count > MAX_REQUESTS
+      total_destroyed += Request.order(id: :desc).offset(Request.count - MAX_REQUESTS).destroy_all.length
```
joshmcarthur: A couple of changes to suggest:

1. You maybe want to order by `created_at`
2. Can't you just `offset(MAX_REQUESTS)`

app/models/sign.rb
elspeth-rabid: @joshmcarthur - I am not sure doing a block/yield type method to record time will be tidy, given that the Request record needs to know the count of results as well.
I tried refactoring but I think this is kind of the clearest.

Br3nda:
I deleted some rows from the database using the heroku console - so prod won't break today.

PR #74, opened 2017-09-12
Fix/staging security

calebkeene:
This is to ensure that the staging site is not indexed by search engines, and add http basic auth.


app/views/layouts/application.html.haml
```diff
@@ -5,6 +5,8 @@
 /[if IE 9 ]    <html lang="en" class="no-js ie9">
 <!--[if (gt IE 9)|!(IE)]><!--> <html lang="en" class="no-js"> <!--<![endif]-->
 %head
+  - if staging_env?
+    %meta{ name: "robots", content: "noindex" }/
```
Br3nda: is the `/` needed? It's not in the other meta in this file

calebkeene:
I'm still going to merge this, because the codeclimate complaints are requesting single quotes in place of double quotes when there is variable interpolation (but the 'rabid way' is double quotes always)

app/views/layouts/application.html.haml
calebkeene: @Br3nda I don't think it's needed (the browser will automatically add it if it's missing) but I think it's best practice to put a trailing slash if it's a self-closing tag, for semantic reasons

PR #79, opened 2017-09-27
Fix/refactors

calebkeene:
Some refactors


Br3nda:
if you run rubocop in your editor, and overcommit withing git, it'll catch most of these linting as you type and before you push.

Gemfile
```diff
@@ -2,15 +2,14 @@ source 'https://rubygems.org'
...
-gem 'rails_12factor'
 
 # Use Postgresql as the database for Active Record
 gem 'haml'
 gem 'jquery-rails'
 gem 'jquery-ui-rails'
 gem 'mini_magick', '~> 3.6.0'
-gem 'nokogiri'
-gem 'pg'
+gem 'nokogiri', '~> 1.8.1'
```
Br3nda: nokogiri gets a few security advisories lately -- if we pin it, deppbot won't send up security upgrades anymore.

app/controllers/application_controller.rb
```diff
@@ -52,10 +52,6 @@ def respond_with_json_or_redirect(object, redirect = nil)
     end
   end
 
-  def load_search_query
-    # @query = session[:search][:query] if session[:search].present? && session[:search][:query].present?
-  end
-
```
Br3nda: :sparkles: !!

app/controllers/items_controller.rb
```diff
@@ -23,35 +24,28 @@ def create # rubocop:disable Metrics/AbcSize, MethodLength, Metrics/PerceivedCom
...
-      flash[:notice] = nil
-      flash[:error] = nil
-      render json: @item
-    else
-      respond_with_json_or_redirect(@item)
-    end
+    return respond_with_json_or_redirect(@item) unless request.xhr?
+
+    flash[:notice] = flash[:error] = nil
+    render json: @item
```
Br3nda: this is waaaaay better! thanks!

app/controllers/sign_image_controller.rb
```diff
@@ -21,13 +21,13 @@ def filename_param
   end
 
   def width_param
-    Integer(sign_image_params[:width])
+    sign_image_params[:width].to_i
```
Br3nda: choice as.

Gemfile
calebkeene: oh but with the squiggle-rocket operator the minor-minor version is unconstrained, so it can freely upgrade to any version that is `1.8.x` where `x` - so the patch version is still covered ✅

calebkeene: I think of `~>` as meaning "equal to or greater than the last digit"

app/controllers/items_controller.rb
calebkeene: ✨ 👾 ✨

spec/controllers/vocab_sheets_controller_spec.rb
```diff
@@ -1,38 +1,113 @@
...
+      it 'displays an error flash message' do
+        invalid_request
+        expect(flash[:error]).to eq I18n.t('vocab_sheet.delete_failure')
+      end
+    end
+
+    it 'redirects to the homepage' do
+      # it redirects to the homepage for valid or invalid requests
+      [valid_request, invalid_request].each do |request|
+        request
```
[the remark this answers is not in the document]
elspeth-rabid: This is definitely a cop you could disable in the spec folder

PR #83, opened 2017-10-04
remove banner from application html

keyaral:
- > Remove partial 

-> Remove code from application.html


Br3nda:
That code climate coverage thing is broken again. Let's turn it off?

joshmcarthur:
@Br3nda it looks like at least some of the Codeclimate checks are still running, possibly just not the overall feedback? I think it's useful to have on, but maybe we could make the `codeclimate/coverage` optional rather than required?

PR #87, opened 2017-10-13
HOTFIX: Work around DNS resolve issues by adding Host header to Freelex requests

joshmcarthur:
Issue:

VUW had freelex.nzsl.vuw.ac.nz and nzsl-assets.vuw.ac.nz CNAME'd to nzsl.vuw.ac.nz. This was a nice efficiency, but when nzsl.vuw.ac.nz was updated to point to Heroku, those CNAMEs followed suit, and stopped pointing where they should (the VUW server that hosts the old Rails app and Freelex).

Solution:

Direct requests to Freelex to the IP address, not the hostname, and add the `Host` header to outgoing server-sent requests to trick that server into picking up the correct virtual host. 

Ongoing issues:

Videos are loaded directly from Freelex client side. I know of no way to trick the browser into adding the host header, and without this, it does fail to resolve. I will experiment with a workaround using a symlink in the old Rails app to see if the videos can become accessible.


joshmcarthur:
image

This is the main visual impact - I have replaced videos with some fixed height text - the height was necessary otherwise layout is affected.

eoinkelly (approved):

ben-rabid (approved):

PR #94, opened 2017-10-17
Vocab Sheet fix for prod

KatieWright26:


KatieWright26:
@elspeth-rabid uuuugh do you understand anything about hound?

PR #105, opened 2017-12-08
Release 7 (Deploy to production)

elspeth-rabid:
Requires testing on staging


Br3nda:
There's conflicts with master *again* -- which is a bit concerning.

Br3nda:
i'm hopeful merging #111 will make this proposed branch green.
I've got an approval from Elspeth on there, but requested changes from Ben. 
I think best if rabid people decide if that gets merged today.

PR #108, opened 2017-12-12
Show search results even if there's a 403 error on image loading

ben-rabid:
Micky reported that searching for "hot" wasn't showing any results on https://www.nzsl.nz/signs/search?utf8=%E2%9C%93&s=hot&hs=&l=&lg=&tag=&usage=

The issue I have found is the Hot water beach sign causing a 403 error which seems to stop foundation triggering the call to show the results.

I have removed the "show if only one" limit to show the search results regardless of count - this does occasionally mean they look a bit uneven, but it's better than no results.


app/assets/javascripts/foundation-equalizer.js
```diff
@@ -1,11 +1,10 @@
 $(document).ready(function() {
-    if ($("#search-container").children().length === 1) {
+    if ($("#search-container").children().length) {
```
KatieWright26: Remove conditional otherwise equalizer wont run

ben-rabid:
This looks like a bigger PR than it is because I indented the JS - all that's new is the removal of `console.log` and the on `window` block

ben-rabid:
@Br3nda do you know what hound is annoyed about?

Br3nda:
the hound error should go away once #110 is merged

PR #109, opened 2017-12-13
Clear Vocab Sheets that are more than one month old.

KatieWright26:
[See zendesk ticked here](https://rabidtech.zendesk.com/agent/tickets/205)

Added dependent to VocabSheet's has_many relationship to ensure that associated items are also removed if a vocab sheet is deleted.


lib/tasks/clear_vocab_sheets.rake
```diff
@@ -0,0 +1,12 @@
+namespace :vocab_sheets do
+  MAX_DAYS = 31
+
+  desc 'Clears all vocab sheets that are more than one month old'
+  task clear: :environment do
+    total_destroyed = 0
+    total_destroyed += VocabSheet.where('created_at < ?', MAX_DAYS.days.ago).destroy_all.length
```
KatieWright26: Might be smarter to change this to 'updated_at < ?' ?

Br3nda: is this was a class method on the VocabSheet model instead, you could write some nice tests.

app/models/vocab_sheet.rb
```diff
@@ -1,7 +1,7 @@
 ##
 # A "sheet" of items (signs) saved by a user
 class VocabSheet < ActiveRecord::Base
-  has_many :items
+  has_many :items, dependent: :destroy
```
ben-rabid: :+1:

lib/tasks/clear_vocab_sheets.rake
```diff
@@ -0,0 +1,12 @@
...
+
+    puts "A total of #{total_destroyed} vocabulary sheets were deleted, #{VocabSheet.count} remaining"\
```
ben-rabid: I expect rubocop will get grumpy at this puts?

ben-rabid (changes requested):
I think the `updated_at` change is essential

ben-rabid (approved):

app/models/vocab_sheet.rb
```diff
@@ -1,9 +1,14 @@
...
 
   def includes_sign?(sign_id)
     items.any? { |i| i.sign_id == sign_id }
   end
+
+  def self.clear_vocab_sheet
```
elspeth-rabid: This is pretty 🐘 but I would call it `delete_old_vocab_sheets`

app/models/vocab_sheet.rb
```diff
@@ -1,9 +1,14 @@
...
+    max_days = 31
```
elspeth-rabid: Perhaps max days could be a constant at the top of the model?

MAXIMUM_AGE_IN_DAYS = 31

app/models/vocab_sheet.rb
elspeth-rabid: The pluralisation is the main thing, `clear_vocab_sheets` is good too

spec/models/vocab_sheet_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+RSpec.describe 'VocabSheet', type: :model do
+  describe '#clear_vocab_sheet' do
+    it 'only deletes a vocab sheet older than 31 days' do
+      FactoryBot.create_list(:vocab_sheet, 5)
+      vocab_sheet = FactoryBot.create(:vocab_sheet, created_at: 1.year.ago, updated_at: 1.year.ago)
```
elspeth-rabid: I would make the old record as close to 31 days as possible to test the edge case - like: `31.days.ago - 1.minute` or `32.days.ago`

spec/models/vocab_sheet_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+      FactoryBot.create_list(:item, 3, vocab_sheet_id: vocab_sheet.id)
+
```
elspeth-rabid: It seems you are testing 2 things here, so it would be great to separate them into 2 tests.

1. `clear_vocab_sheet` deletes the sheet that is more than 31 days old and not the others - checking both the number of records and that the remaining ones have `updated_at` less than 31 days ago.

2. That when the vocab sheet is deleted, the items are too.

This is fine, it is just not super obvious what you are checkin and why based on the `it` block statement alone.

spec/rails_helper.rb
```diff
@@ -51,4 +51,15 @@
   # The different available types are documented in the features, such as in
   # https://relishapp.com/rspec/rspec-rails/docs
   config.infer_spec_type_from_file_location!
+
```
elspeth-rabid: 👍

elspeth-rabid (approved):
Nice! I made a few comments on naming and testing stuff but nothing to hold up merging.

spec/models/vocab_sheet_spec.rb
Br3nda: i love that there are tests now.... so we can be more confident future devs won't make a mistake and delete all data.

lib/tasks/clear_vocab_sheets.rake
```diff
@@ -0,0 +1,5 @@
+namespace :vocab_sheets do
+  task clear: :environment do
+    VocabSheet.clear_vocab_sheet
```
Br3nda: the name doesn't quite capture what it does.
maybe `purge_old_sheets!` <-- :elephant:

Br3nda: aha.. Elspeth said the same thing :-D

app/models/vocab_sheet.rb
Br3nda: this could be in config. Then we have control of it without changing code and doing a release. We could even turn it off.

Br3nda (approved):
2x :elephant: comments.

Br3nda:
So, is the database cleaner addition to make it run faster?

KatieWright26:
Nah, I just noticed when I ran my tests multiple times that my FactoryBot
list of VocabSheets weren't being cleared, so installed the db cleaner to
make sure all tests start fresh.

PR #115, opened 2017-12-14
Remove my previous change to how search results are shown

ben-rabid:
Sadly it didn't fix the issue when we got to a slower server.

Micky has said for now she will fix up the Freelex data and we can leave it there.


elspeth-rabid (approved):
Since you are just reverting I approve but let me know if you are getting weird behaviour for the search results, Katie and I know Foundation Equalizer pretty inside out.

Br3nda (approved):

PR #133, opened 2018-01-25
Auto corrected by following Capybara/FeatureMethods

Br3nda:
Auto corrected by following Capybara/FeatureMethods

Click [here](https://awesomecode.io/projects/58/ruby_configs/14960) to configure it on awesomecode.io


Br3nda:
i really like bots

PR #135, opened 2018-01-25
Bug/fix advanced search layout

KatieWright26:
Fixing bug from [here](https://www.pivotaltracker.com/story/show/152420166). For time being I have added scrolling functionality so that users can still navigate the search options.

The issue is arising from the content container having `overflow: hidden` set. Unfortunately this affects how the vocab sheet is displayed, and turning it off causes VS elements to sit above the footer. The vocab sheet really needs a refactor that is probably out of scope for SLA.

Looked at applying Foundation equalizer to VS; but doesn't appear to work when haml is spread across different files. Experimented with jquery loading the VS at the same time as the search results; but there are too many edge cases to account for (is the search container on the page, does it contain results, is there a flash banner on the page, do the elements in the VS overflow the container and require a scroll bar etc.


app/assets/javascripts/foundation-equalizer.js
```diff
@@ -5,7 +5,6 @@ $(document).ready(function() {
     $("#search-container").on("postequalized.zf.equalizer", function() {
       $results = $(".link_and_drawing_container");
       $results.each(function(){
-        console.log("found");
```
Br3nda: :sparkles:

Br3nda:
I can't see that issue, because the pivotal tracker is private.

I've checked out the branch and tested in Firefox and it looks good to me.

KatieWright26:
Oops, ok I'll write proper discriptions in future; this was to fix the
problem of the advanced search menus being cut off when they overflow the
main content container. Also, I can't merge this; I don't have write access
on this project.

Br3nda:
@KatieWright26 i sent you an invite to the organisation earlier this morning

KatieWright26:
@Br3nda Yup I accepted that, but I think you need to explicitly give me write privileges as well: 

screen shot 2018-01-25 at 3 56 05 pm

PR #138, opened 2018-01-30
Upgrade rails to version 5

### LINKED ISSUE #132: Upgrade to rails 5
TODO

Useful instructions are here https://hashrocket.com/blog/posts/how-to-upgrade-to-rails-5

KatieWright26:
closes #132


Br3nda:
This is a beautiful thing you've done @KatieWright26

Gemfile
```diff
@@ -1,14 +1,14 @@
 source 'https://rubygems.org'
 ruby '2.3.6'
 
-gem 'rails', '~> 4.2.6'
+gem 'rails', '~> 5.0.0'
```
Br3nda: Why not all the way to 5.1.4?

app/controllers/vocab_sheets_controller.rb
```diff
@@ -12,7 +12,7 @@ def show
   end
 
   def update
-    @sheet.name = params[:vocab_sheet][:name]
+    @sheet.update_attributes!(vocab_sheet_params)
```
Br3nda: I right now think it will never reach line 16 if the validations fail - the code will throw an exception on line 15

We could instead do `if @sheet.update(vocab_sheet_params)`

app/controllers/vocab_sheets_controller.rb
```diff
@@ -12,7 +12,7 @@ def show
...
     if @sheet.save
       flash[:notice] = t('vocab_sheet.sheet.update_success')
     else
```
Br3nda: With these changes, the `else` can never be reached.

app/controllers/vocab_sheets_controller.rb
```diff
@@ -37,6 +37,10 @@ def destroy
 
   private
 
+  def vocab_sheet_params
+    params.require(:vocab_sheet).permit(:name)
+  end
```
Br3nda: 👍

app/controllers/vocab_sheets_controller.rb
KatieWright26: Like this?

Br3nda:
On staging now

KatieWright26:
@Br3nda the search on staging is broken by this ( I think), do you have the ability to revert?

Br3nda:
i can make staging back like the master branch, yes

PR #148, opened 2018-02-27
Bug/videos not playing in safari

KatieWright26:
Videos in the sign show pages were not playing in Safari only. Discovered that the custom play button was preventing default behaviour across all videos, not just those with the custom play button overlaid.

Added a conditional to check for custom play button. Tested in Chrome, Safari and Firefox.


KatieWright26:
Branch contains content from rabid staging fork that should not be merged .

PR #158, opened 2018-04-09
Bugfix/zendesk 296 body location in advanced search not working

Rabid-Dan:
A bug is currently effecting the "body", "arm", and "hand" filters in the body location advanced search options menu.
Selecting these and searching for any term adds the "top of head", "eyes" and "nose" options as filters instead. This is due to the javascript looking for the presence of a `main_image` class on these particular search option while assigning query params but not finding one. 

It doesn't find one because that class is dynamically assigned in `search_helper.rb` and the lack of any spacing between the default classes and the dynamically assigned class is causing `main_image` to be chained onto the end of the last default class (i.e instead of the classes assigned being `image rounded main_image` they are `image roundedmain_image` meaning it has neither `rounded` or `main_image` and instead has `roundedmain_image`).

This fix adds a space to the end of the default class string and a comment explaining why it is necessary. It also lightly refactors the `classes_for_sign_attribute` method.
I've chosen to add the beginnings of a search helper spec as well with some basic value assertions for the current cases in that method. 
Hopefully this will provide some extra context in the event that someone in the future decides to ignore the commenting and remove the space from the default classes string.


app/helpers/search_helper.rb
```diff
@@ -183,12 +183,12 @@ def value_for_sign_attribute(number, attribute, main)
...
-    classes = 'image rounded'
-    if main
-      classes += 'main_image'
-    elsif attribute == :handshape
-      classes += 'transition'
-    end
+    # a space is required after the base class names below to ensure that they are
+    # properly separated from other classes assigned in the conditionals.
+    classes = 'image rounded '
+    classes += 'main_image' if main
```
elspeth-rabid: Surely a space after each class? Or push the classes onto an array and join with spaces at the end of the method?

spec/helpers/search_helper_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
```
elspeth-rabid: ❇️

app/helpers/search_helper.rb
```diff
@@ -183,12 +183,12 @@ def value_for_sign_attribute(number, attribute, main)
...
+    classes += 'transition' if attribute == :handshape
+
```
Br3nda: is `main_imagetransition` a class? You'd get that if main && attribute== :handshape

How about assembling an array, and joining it with spaces?
```
classes = ['image', 'rounded']
classes << 'main_image' if main
classes << 'transition' if attribute == :handshape
classes.join(' ')
```

spec/helpers/search_helper_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    context 'when provided with an :handshape attribute' do
+      it 'returns the transition class as part of the response string' do
+        expect(helper.send(:classes_for_sign_attribute, :handshape, false)).to eq('image rounded transition')
+      end
+    end
+
+    context 'when provided with a false value for main and a attribute not matching :handshape' do
+      it 'returns the default class string' do
+        expect(helper.send(:classes_for_sign_attribute, nil, false)).to eq('image rounded ')
+      end
```
Br3nda: how about splitting the classes by ' ' and then checking that image and rounded were in there?
Then this test is less likely to break in future css/classname changes.

spec/helpers/search_helper_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    context 'when provided with an :handshape attribute' do
+      it 'returns the transition class as part of the response string' do
+        expect(helper.send(:classes_for_sign_attribute, :handshape, false)).to eq('image rounded transition')
+      end
+    end
+
+    context 'when provided with a false value for main and a attribute not matching :handshape' do
+      it 'returns the default class string' do
+        expect(helper.send(:classes_for_sign_attribute, nil, false)).to eq('image rounded ')
```
Br3nda: how about splitting the classes by ' ' and then checking that `image` and `rounded` were included in there?
Then this test is less likely to break in future css/classname changes.

Br3nda: there might be a smart railsy-way to check for classes on an element.

app/helpers/search_helper.rb
Rabid-Dan: oh yeah I didn't consider both of them being added. Thats a nice way of doing it! I'll update the branch now

PR #167, opened 2018-06-24
Support/update sprockets

louise-r-blue:


louise-r-blue:
Regarding the last commit. I had installed imageMagick to enable images to be loaded when the site loaded which has updated the ruby and therubyracer gem.

Br3nda:
Thanks for this. We used to have @deppbot but that bots has gone dead.

LGTM, then set up renovate bot to automate patching from now on. the robots will always be faster at this.

PR #185, opened 2018-08-06
Feature/vocab scroll bar

ghostPepperWonton:
- Fixed a line in the readme
- Styled the side column
- Removed the background image from the container and moved it into the styles for the side column
- Changed the positioning so that CSS does the calc instead of JS
- Changed out the delete icon for something that has a transparent background


app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -1,11 +1,20 @@
 /* Vocab bar background */
 .vocab_sheet_background {
-  background: right top repeat-y;
```
Rabid-Dan: would be good to remove this if we can confirm browser support requirements before merging

Rabid-Dan:
So it looks like this was maybe branched off the base repo rather than the rabid fork.
Also I'm seeing some funky display stuff (images below)

Side bar running above title banner
screen shot 2018-08-07 at 11 33 53 am

app/assets/stylesheets/_vocab_sheet.scss
ghostPepperWonton: I was able to run this through BrowserStack by the way, looks like we're good for IE 10+ like we saw on caniuse, so that's good unless we need IE <= 9 support.

ghostPepperWonton:
@Rabid-Dan Your screenshot issue has been addressed.

Rabid-Dan (approved):

kasperite (approved):

kasperite:
would be good to have an attached screenshot in description, otherwise :shipit:

PR #186, opened 2018-08-06
Accurately reflect changes to vocab sheet layout in the view

Rabid-Dan:
Currently a combination of extra margins and fixed widths for items in the vocab sheet are causing layouts to present inaccurately (e.g selecting the 3 column layout will instead present two columns).

To fix this I've opted to add some flex properties to the vocab items container and remove the fixed width from items themselves. I've also added a max width to the images on each item to stop a visual error where wide images would push the container width out slightly.

Images below:

screen shot 2018-08-06 at 3 22 21 pm

screen shot 2018-08-06 at 3 22 32 pm


ghostPepperWonton (approved):

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -1,7 +1,7 @@
-%li.vocab_sheet_item.vocab_sheet_page_item.shadow.rounded{ style: "width:#{vocab_sheet_item_width}px;height:#{vocab_sheet_item_height}" }
+%li.vocab_sheet_item.vocab_sheet_page_item.shadow.rounded
   = hidden_field_tag "item[#{vocab_sheet_item.id}][id]", vocab_sheet_item.id, class: 'item_id'
   = link_to sign_url(vocab_sheet_item.sign_id), { class: 'drawing vocab_sheet_drawing', style: "line-height:#{vocab_sheet_image_height}px;height:#{vocab_sheet_image_height}px" } do
-    = image_tag "signs/#{vocab_sheet_image_width}-#{vocab_sheet_image_height}/#{vocab_sheet_item.drawing}"
+    = image_tag "signs/#{vocab_sheet_image_width}-#{vocab_sheet_image_height}/#{vocab_sheet_item.drawing}", style: "max-width:#{vocab_sheet_image_width - 7}px"
```
fractalcactus: is it possible to avoid inline styling?

calebkeene: seconded, this is undesirable

README.md
```diff
@@ -6,7 +6,7 @@ New Zealand Sign Language Dictionary consists of 2 major units
 
 [![Build Status](https://travis-ci.org/rabid/nzsl-online.svg?branch=master)](https://travis-ci.org/rabid/nzsl-online)
 [![Code Climate](https://codeclimate.com/github/rabid/nzsl-online/badges/gpa.svg)](https://codeclimate.com/github/rabid/nzsl-online)
-[![Test Coverage](https://codeclimate.com/github/rabid/nzsl-online/badges/coverage.svg)](https://codeclimate.com/github/rabid/nzsl-online/coverage)
+[![Test Coverage](https://codeclimate.com/github/rabid/nzsl-online/badges/coverage.svg)](https://codeclimate.com/github/rabid/nzsl-online/coverage)
```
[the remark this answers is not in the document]
Br3nda: This needs changing to be the ODNZSL instead of Rabid.

app/views/vocab_sheets/_vocab_sheet_item.html.haml
Rabid-Dan: I think the reason all the inline styling is done here is because it uses helper methods to evaluate height and width values based on the page layouts.
I'm pretty sure I could change it to work through class assignment rather than style assignment from those helpers but it looks like it would add an extra chunk of work onto this branch.
What do you reckon?

README.md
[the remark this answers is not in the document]
Rabid-Dan: thanks for catching this

app/views/vocab_sheets/_vocab_sheet_item.html.haml
Br3nda: It has a higher connaisance as an inline style

Br3nda: http://connascence.io/

Rabid-Dan:
@Br3nda @fractalcactus @calebkeene I've resolved the inline styling comments by using class assignment instead. I've been told that we're actively moving away from haml as well so I've taken the opportunity to convert the view into an html.erb partial instead.

PR #187, opened 2018-08-06
Update/test conformance

ghostPepperWonton:
- Byebug keeps adding its history to the repo, just ignoring it in the gitignore.
- All of the failing tests were related to linked images that have moved in Freelex, so I just moved them to an updated image.
- This also requires an S3 bucket URL in your secrets which I haven't been able to work out with ops yet, so that might fail depending on what you've been able to set up locally.
- There's also an unwritten tests for the User model that isn't in use.


ghostPepperWonton:
This branch ended up including a whole bunch of changes from the wrong branch. I've made a new branch and a new pull request. Disregard this.

PR #196, opened 2018-08-08
`let` was causing failing build

ghostPepperWonton:
Using `var` and trying to ignore Code Climate build


Rabid-Dan (approved):

Br3nda:
Need to tell eslint we aren't using es6

PR #198, opened 2018-08-09
Auto corrected by following Javascript semi

Br3nda:
Auto corrected by following Javascript semi

Click [here](https://awesomecode.io/projects/58/javascript_configs/1117) to configure it on awesomecode.io


Br3nda:
closed. I don't konw what  app/assets/javascripts/dd_belatedpng.js is, but it looks like it shouldn't be edited

PR #199, opened 2018-08-09
Auto corrected by following Javascript semi

Br3nda:
Auto corrected by following Javascript semi

Click [here](https://awesomecode.io/projects/58/javascript_configs/1117) to configure it on awesomecode.io


nzcodarnoc (approved):
There's always a tricky semicolon

PR #202, opened 2018-08-13
Bugfix/update devise

Rabid-Dan:
Staging is currently throwing 503's and after a bit of trouble shooting and a quick google it looks like its a problem that the current version of devise is having with ruby 2.5 (see issue here: https://github.com/plataformatec/devise/issues/4736)

Theres a fix bundled into the devise 4.4 so this bugfix updates to the newer version.


ben-rabid (approved):

louise-r-blue (approved):

Br3nda:
I think we need some feature specs, so we catch that next time too.

PR #208, opened 2018-08-14
Vocab sheet layout options update

Rabid-Dan:
Original story was rejected because the two column layout is too tall and now bumps the copyright notice onto another page in print views.

I suspect this may have been happening for a while but its related enough to the layout options story that an update of this size feels fine.

This branch adds a small amount of print specific styling for that column layout to reduce image sizes in order to keep everything on a single page (image below:).

screen shot 2018-08-14 at 12 08 52 pm


Br3nda (approved):
works great in firefox

louise-r-blue:
Great looks very tidy

ghostPepperWonton (approved):
Looks good! Will potentially help with the font-sizing being dynamic later on too with a set bounding area.

louise-r-blue (approved):
Very tidy solution

PR #232, opened 2018-08-19
Feature/enhance default placeholders

louise-r-blue:
This is the updated placeholder which keeps as close to wireframe design and works in Safari, Chrome and Firefox but please double check. (Thanks Brenda for spotting that earlier)
chrome-placeholder

Originally I has tried to keep the placeholder content to as close to possible to the design story. However the search results container displays with different heights. This is more noticeable between the first row and any consecutive rows. This meant the placeholder content was not consistent from row to row. This would be only noticeable if the user was accessing search results on a very slow network but I decided to simplify the placeholder content to the bare minimal.
Therefore the placeholder content is very skeletal and consists of 2 pale blue rectangles - one for the video and one to represent the gloss text.
placeholder


app/assets/javascripts/search-result-load-in.js
```diff
@@ -1,5 +1,5 @@
 $(document).ready(function() {
-  var searchResult = 0; // eslint-disable-line no-var
+  let searchResult = 0; // eslint-disable-line no-var
```
Rabid-Dan: I think this will need to change back to var, pretty sure @danieljauch had issues with let

app/assets/stylesheets/placeholder.scss
```diff
@@ -0,0 +1,15 @@
+/* blue rectangle at the top covering the video with color at a height of 100% */
+.video-container.placeholder {
+  background-image: linear-gradient($pale-blue 100%, transparent 0);
+  background-repeat: no-repeat;
```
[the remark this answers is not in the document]
ghostPepperWonton: If you haven't seen this one before, CodeClimate wants CSS to be alphabetical, so that's a way you can avoid more of these warnings.

Br3nda:
in firefox, I'm seeing a blue rectangle on each card

image

app/assets/javascripts/search-result-load-in.js
```diff
@@ -1,11 +1,11 @@
 $(document).ready(function() {
   var searchResult = 0; // eslint-disable-line no-var
-
   $('.search-results__card').each(function() {
     $(this)
       .css({
         transitionDelay: (searchResult * 0.05) + 's',
       })
+      .removeClass('search-results--placeholder')
       .addClass('show-card');
```
ghostPepperWonton: It's probably better to take line 8 and move it into a `setTimeout` after the `transitionDelay`. The way you have it now, it removes the placeholder class before it's shown.

app/assets/stylesheets/placeholder.scss
```diff
@@ -0,0 +1,25 @@
+.video--placeholder {
+  background-image: linear-gradient($pale-blue 100%, transparent 0); // blue rectangle at the top covering the video with color at a height of 100%
```
ghostPepperWonton: Right now, this is covering 100% of the area with the `$pale-blue` color, so instead of `background-image` with a `linear-gradient`, you should use `background-color`.

app/assets/stylesheets/placeholder.scss
```diff
@@ -0,0 +1,25 @@
...
+    0 70%, // middle line
+    0 75%, // bottom line
+    100% 100%; // bottom right square
+  background-repeat: no-repeat;
+  background-size: // background-size maps to background-position
+    80% 20%, // top line width + height
+    80% 15%, // middle line
+    80% 15%, // bottom line
+    50% 25%; // bottom right square
+}
```
ghostPepperWonton, line 18: Percentage-based gradients will shift as the size of the content comes in, so it's better to put the `background-color` on the individual elements you're trying to put this color behind. Based on this morning, I think you're already doing this, but just wanted to make sure we aren't putting in a style that won't work long-term.

app/assets/javascripts/search-result-load-in.js
louise-r-blue: mme I did have after but in a SetTimeout. Will try this

app/assets/stylesheets/placeholder.scss
louise-r-blue: the 100% is the height of the video-container area only not the total container.

app/assets/stylesheets/placeholder.scss
louise-r-blue: There are a multitude of ways to do this and so I went for this for now at least. The rows are not the same height and currently the elements with-in the containers are of un-equal heights and not positioned evenly so using a % on the background meant I would get it as close to the individual height as device size changes etc: and to match the placeholder wireframe for consistency. I understand what you are saying I don't think it will be consistent using the container elements as is right now at least. I think there is a story made since the load time changed that may help address this issue. I also thought that extracting and putting the placeholder style in a seperate file may be better in long term?

louise-r-blue: I have added a setTimeout that removes the placeholder style and changed the video-container placeholder with a style in the _video.scss file

ghostPepperWonton (approved):
Thanks for dealing with my nit-pickiness :)

PR #253, opened 2018-08-21
Feature/print footer

ghostPepperWonton:
I made the application, header, and footer partials into ERB files, and separated out the show and print views for vocab sheets to make styling and layout simpler.


app/assets/stylesheets/print.scss
```diff
@@ -201,3 +206,9 @@ h1, .h1 {
   width: 690px;
   padding: 0;
 }
+
+@media only print {
+  .page__header {
+    display: none;
```
louise-r-blue: Just so I understand does the page___header class relate to the nzsl header (New Zealand sign language dictionary ) for the print pages/s

louise-r-blue (approved):

PR #256, opened 2018-08-21
Auto corrected by following Javascript no-lonely-if

Br3nda:
Auto corrected by following Javascript no-lonely-if

Click [here](https://awesomecode.io/projects/58/javascript_configs/1097) to configure it on awesomecode.io


Br3nda:
OMG i swear i turned this off 4 times now

PR #259, opened 2018-08-21
Removed insecure use of `open` in sign.rb

Br3nda:
* Removed the brakeman ignore file
* added faraday
* fixes our security mess with the use of `open` to do http requests.
* moves Image Processor from models to services, because it is a service. Then brakeman stops thinking it's a model
* Fix code coverage reporting


Br3nda:
i suspect the `k` variable needs escaping in here too, but would love a second opinion

```
      if v.is_a?(Array)
        v.each { |ea| query_string << "#{k}=#{CGI.escape(ea.to_s)}" if ea.present? }
      elsif v.present?
        query_string << "#{k}=#{CGI.escape(v.to_s)}"
      end
```

Br3nda:
it's not following the http 302 on request for a random one. I'll fix

app/models/sign.rb
```diff
@@ -76,26 +75,20 @@ def self.search(params)
...
-
-  def self.url_for_search(query)
+  def self.uri_for_search(query)
     # The handling of arrays in query strings is different
     # in the API than in rails
     return SIGN_URL unless query.is_a?(Hash)
+    '?' + _query_string_for_search(query).join('&')
+  end
+
+  def self._query_string_for_search(query)
```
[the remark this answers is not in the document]
Br3nda: I've reduced this from 11.

Br3nda:
I've moved the ImageProcessor to services folder, so brakeman doesn't think it's a model.

app/models/sign.rb
```diff
@@ -76,26 +75,20 @@ def self.search(params)
   end
 
   def self.xml_request(params)
-    xml_document = nil
-    url = url_for_search(params)
-    time = Benchmark.measure do
-      xml_document = Nokogiri::XML(open(url))
-    end
+    xml_document = Nokogiri::XML(http_conn.get(uri_for_search(params)).body)
```
elspeth-rabid, line 78: Yay! No more slow queries :-)

app/models/sign.rb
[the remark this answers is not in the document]
elspeth-rabid: What is the leading underscore convention?

app/models/sign.rb
```diff
@@ -104,6 +97,13 @@ def self.url_for_search(query)
         query_string << "#{k}=#{CGI.escape(v.to_s)}"
```
elspeth-rabid, line 97: Well... the keys are not user provided but easily tampered with... I'd say escape em!

spec/models/sign_spec.rb
```diff
@@ -52,12 +52,12 @@
   end
 
   describe '.random' do
-    it 'retrieves two random signs' do
-      sign_one = Sign.random
-      sign_two = Sign.random
-      expect(sign_one.is_a?(Sign)).to eq(true)
-      expect(sign_two.is_a?(Sign)).to eq(true)
-      expect(sign_one.id).not_to eq(sign_two.id)
+    describe 'retrieves two random signs' do
```
elspeth-rabid, line 55: Nicer!

app/models/sign.rb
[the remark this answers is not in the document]
Br3nda: oops. I've been coding in python more than ruby. It's used to mark a method as private.

Ruby has it's own way.

I think some of these class methods could go into a service instead. It's pretending to be ActiveRecord, but it's not.

app/models/sign.rb
Br3nda: if you follow it back up through the stack, the keys do come from `params`

app/models/sign.rb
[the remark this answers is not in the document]
elspeth-rabid: I do agree, we will certainly be refactoring model and controller stuff into services when we do touch files.
I think the underscore should possibly go.

app/models/sign.rb
elspeth-rabid: Yeah, but with GET right? So the params are sent in the URL and can be tampered. Or am I missing the point? Ha ha.

Br3nda:
They're sent in a http GET to FreeLex - so yes. They could be something
malicious.

app/models/sign.rb
Br3nda: They're sent in a http GET to FreeLex - so yes. They could be something malicious.

ghostPepperWonton (approved):

PR #261, opened 2018-08-21
Feature/print footer

ghostPepperWonton:


louise-r-blue (approved):

app/assets/stylesheets/application.scss
```diff
@@ -95,7 +95,7 @@ a.logo {
   border:0;
 }
 
-.print_logo { display:none; }
+.logo--print { display:none; }
```
Rabid-Dan, line 98: ⭐️ 🅱️ **E** Ⓜ️ ⭐️

app/helpers/vocab_sheet_helper.rb
```diff
@@ -34,11 +34,11 @@ def vocab_sheet_offset_multiple
       if @size == 5
         15
```
Rabid-Dan, line 35: This is a pretty big 🐘 but now might be a good time to refactor some of the vocab_sheet_pages method below. It took me a little bit to figure out what was going on in there so maybe some wrapper methods around some of the logic might be a good idea?

Rabid-Dan (approved):

app/helpers/vocab_sheet_helper.rb
ghostPepperWonton: Yeah, the `vocab_sheet_pages` method isn't actually used with the current iteration, but it has potentially good uses, so I've left it for the moment.

PR #264, opened 2018-08-22
Fixes rubocop which was ignoring all files

Br3nda:
I haven't worked out why, but when these now-removed lines are included, rubocop checks zero files.


ghostPepperWonton (approved):
My code editor has been complaining about this for a week :P

PR #268, opened 2018-08-22
Feature/add to vocab sheet from search

ghostPepperWonton:


Br3nda:
I find the ERB harder to read - it also doesn't catch nesting errors and invalid markup the way haml does. Why translate the haml over to ERB?

ghostPepperWonton:
@Br3nda Rabid uses ERB as a standard at the moment, so we've been switching templates over as we use them. This seems like a bigger discussion we should have with you though.

.codeclimate.yml
```diff
@@ -0,0 +1,6 @@
+engines:
+  eslint:
+    enabled: true
+    checks:
+      no-var:
+        enabled: false
```
Rabid-Dan, line 6: 🌟

louise-r-blue (approved):
Looking good

app/assets/javascripts/search-vocab.js
```diff
@@ -0,0 +1,55 @@
+$(document).ready(function() {
+  $('.add-to-vocab-btn').click(function(e) {
+    e.preventDefault();
+    var signId = $(this).attr('data-sign-id');
+
+    var successNotice =
```
Rabid-Dan: How would you feel about having these elements always present in the sidebar but hidden and just toggling them to show/hide on success/failure?

app/assets/javascripts/search-vocab.js
```diff
@@ -0,0 +1,55 @@
...
+      url: '/vocab_sheet/items/',
+      method: 'POST',
+      data: {
+        sign_id: signId,
+      },
+      headers: {
+        'X-CSRF-Token': $('meta[name="authenticity-token"]').attr('content'),
+      },
+    }).done(function(data) {
+     if ($('.vocab_sheet_bar').css('display') === 'none') {
```
Rabid-Dan: Do you think this and the failure stuff below could get pulled out into separate functions to tidy up the ajax call?

app/assets/javascripts/search-vocab.js
ghostPepperWonton: I kind of prefer this way now that we're using Ajax, so that if some one double-clicks, they get two messages. Thoughts?

app/assets/javascripts/search-vocab.js
ghostPepperWonton: I can do that :)

app/assets/javascripts/search-vocab.js
ghostPepperWonton: This is resolved by the current update :)

app/assets/javascripts/search-vocab.js
```diff
@@ -0,0 +1,56 @@
+$(document).ready(function() {
+  $('.add-to-vocab-btn').click(function(e) {
```
Rabid-Dan: Oh one last thing! I think its probably worth wrapping the all the logic/functions here in a initialisation function and then calling that function at the bottom but **only** if the .add-to-vocab-btn element is present (or some sort of similar presence check). That way we can stop it from trying to run all of this on pages that don't need the functionality or don't feature the class which could potentially throw some console errors.

Rabid-Dan (approved):
⭐️ 🏌️

Br3nda:
@danieljauch that's a pain because we converted them all to haml only last year. There needs to be an advantage, rather than just that Rabid have a standard.

PR #269, opened 2018-08-23
Feature/test text editing

louise-r-blue:
Some very basic styling to get the print view and actual print- out working on a few bare minimal styling rules.


app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -150,15 +150,24 @@
...
 
-.vocab_sheet_page_item .item_name {
+.vocab_sheet_page_item .item_name,
+.vocab_sheet_page_item .item_maori_name {
   margin-bottom: 5px;
+  margin-right: 0px !important;
   position: relative;
   z-index: 11;
+  height: 5em;
+  // white-space: pre;
```
Rabid-Dan: can these commented bits be removed if they're not going to be re-enabled please

app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -297,3 +306,22 @@
 .size_controls .selected_icon {
   background-color: #d2e2f9;
 }
+
+p.print_name {
```
Rabid-Dan, line 305: Are these all print specific styles? If so they should probably go into the print style sheet

app/views/vocab_sheets/_vocab_sheet_item.html.erb
```diff
@@ -1,15 +1,13 @@
-<div class="vocab_sheet_item__column vocab_sheet_item__column--<%=size%>">
+<div class="print_view vocab_sheet_item__column vocab_sheet_item__column--<%=size%>">
```
Rabid-Dan, line 1: This partial is used in both the print and non print views so this new class should probably get removed and attach any related styles to one of the other existing classes in the print style sheet instead.

ghostPepperWonton: Since you're creating this class name, you should create it with BEM practice in mind. Example: `vocab-sheet-item--print-view`

app/assets/stylesheets/_vocab_sheet.scss
louise-r-blue: yes you are right.  Thank you : )

app/views/vocab_sheets/_vocab_sheet_item.html.erb
louise-r-blue: Thank you. Will refactor once I get my head around all the different styles : }

louise-r-blue:
I am deleting this branch as not needed

PR #270, opened 2018-08-23
Feature/vocab add error text

ghostPepperWonton:


Rabid-Dan (approved):

Br3nda:
hang on -- is @danieljauch and @Rabid-Dan different people!

Rabid-Dan:
@Br3nda haha we are and coincidentally we sit next to each other at the same pod here as well, so we are sort of interchangably "The Dans"

louise-r-blue (approved):

PR #272, opened 2018-08-23
Update/loading placeholder

louise-r-blue:
2 examples of throttling (Firefox and Google) to capture the placeholder.

firefox_throttling_good_3g
google_slow_3g

This commit is a follow up from sprint review with Micky the product owner today. The user story NZSL-4 was rejected based on the fact user could not see the placeholder before contents were fully loaded in search page results. 

This commit removes the setTimeout function that removed the class containing the placeholder css and replaces it with the window on 'load' jQuery function which removes the placeholder class once the content on the search results page fully loads.


app/assets/javascripts/search-result-load-in.js
```diff
@@ -3,12 +3,12 @@ $(document).ready(function() {
   $('.search-results__card').each(function() {
     $(this)
       .css({
-        transitionDelay: (searchResult * 0.05) + 's',
+        transitionDelay: (searchResult * 0.10) + 's',
```
ghostPepperWonton: Tiny 🐘 : I don't think we need that trailing zero. You could probably just write it as 0.1. Super small thing.

app/assets/javascripts/search-result-load-in.js
```diff
@@ -3,12 +3,12 @@ $(document).ready(function() {
...
       })
       .addClass('show-card');
-      setTimeout(function() {
-        $('.search-results__card').removeClass('search-results--placeholder');
-      }, searchResult * 50 );
     searchResult++;
   });
+  $(window).on('load', function() {
+    $('.search-results__card').removeClass('search-results--placeholder');
+  });
```
ghostPepperWonton: One small thing here. Dan had me wrap my JavaScript in an if statement that checks if there are elements to assign these sort of things on before assigning them. So what that would look like is something like...
```
  if ($('.search-results__card').length > 0) {
    // the .each() method
    // the window load check
  }
});
```
JavaScript does some bad things with memory if you have methods run on things that don't exist. This will also help make sure all the JavaScript loads faster on pages that _don't_ include this class.

ghostPepperWonton (approved):
Looks great ✨

Br3nda (approved):

PR #273, opened 2018-08-26
Feature/notes print view

ghostPepperWonton:
- Added notes view as an option
- Added `notes` property to `Items`
- Changed icons for vocab sheet layout options
- Created and styled views

screen shot 2018-08-27 at 12 42 54 pm
screen shot 2018-08-27 at 12 44 35 pm


louise-r-blue (approved):

app/views/vocab_sheets/show.html.haml
```diff
@@ -0,0 +1,36 @@
...
+- content_for :head do
+  = stylesheet_link_tag "smoothness/jquery-ui.css"
+
+.vocab-sheet__head
+  .vocab-sheet__controls.vocab-sheet__page-controls
+    = orange_submit_button "vocab_sheet.print", {print: true}
+    = form_tag vocab_sheet_path, method: :delete, class: "" do
+      = orange_submit_button "vocab_sheet.delete"
+    .vocab-sheet__size-controls
+      .vocab-sheet__size-control-box
```
Rabid-Dan, line 12: Why did this get converted from the loop it was in? I think it looked a lot tidier the old way

Rabid-Dan:
nicceee, just the one comment one here

app/views/vocab_sheets/show.html.haml
ghostPepperWonton: I took it out because I was adding the notes view which doesn't have a number, but it totally could be returned to that since it's in a separate container. Doing that now.

app/controllers/items_controller.rb
```diff
@@ -36,6 +36,7 @@ def create # rubocop:disable Metrics/AbcSize, MethodLength, Metrics/PerceivedCom
   def update # rubocop:disable Metrics/AbcSize
     @item.name = params[:item][:name] if params[:item][:name].is_a?(String)
     @item.maori_name = params[:item][:maori_name] if params[:item][:maori_name].is_a?(String)
+    @item.notes = params[:item][:notes] if params[:item][:notes].is_a?(String)
```
Rabid-Dan: After discussing with Daniel we're deciding to use permitted params here instead of individual checks for a more standards based approach

ghostPepperWonton: This is now in :)

app/controllers/items_controller.rb
```diff
@@ -33,10 +30,8 @@ def create # rubocop:disable Metrics/AbcSize, MethodLength, Metrics/PerceivedCom
     end
   end
 
-  def update # rubocop:disable Metrics/AbcSize
-    @item.name = params[:item][:name] if params[:item][:name].is_a?(String)
-    @item.maori_name = params[:item][:maori_name] if params[:item][:maori_name].is_a?(String)
-    if @item.save
+  def update
+    if @item.update(item_params)
```
Rabid-Dan, line 34: ⭐️

Rabid-Dan (approved):

PR #275, opened 2018-08-27
Feature/vocab sidebar flashes

ghostPepperWonton:
This addresses two issues that Micky raised in testing:

- If you add a vocab item to your list from the search, then remove it, it wasn't hiding the column and was duplicating the "empty vocab list" message
- If you delete an item from the list the "Item removed" message was never going away.


Rabid-Dan (approved):
⭐️

louise-r-blue (approved):

PR #277, opened 2018-08-27
Add migration step to procfile

Rabid-Dan:
Heroku currently doesn't run migrations as part of its automatic deployment process when connected with Github, which has just caused us some issues after deploying an update to staging which added a new notes column to the Item model.

They do however provide support for release phase tasks which will run them for us after a successful app build https://devcenter.heroku.com/articles/release-phase (this includes a nice protection where if the release phase task fails for any reason then the build doesnt get deployed).

This update adds a migration release task to the procfile.


ghostPepperWonton (approved):

louise-r-blue (approved):
Nice

PR #278, opened 2018-08-27
Feature/enhance vocab sheet size

louise-r-blue:
NEW---------------------------------------------------------

Since the functionality to expand the textures was converted from js to jquery quite a few issues have arisen that will be addressed in due course.

OLD----------------------------------------------------------

Need some eyes on this please.

Currently I have left the autoExpand textarea file reasonably well commented. The comments can be extracted and added to a further PR if required.

The autocise functions include a check of textarea input scroll height so that when the users vocab sheet page loads or reloads it should display all the text without any obfuscation. Then to keep each row's ( main gloss row and Maori gloss row) height even and consistent for display and print there is a further check determining the max height that all textareas on a row increase height at the same time.

In response to the clients comments about max gloss size I have made the textures both editable, expandable up to a max length of 100 characters - which can be altered if desired.


app/assets/javascripts/autocise_textarea.js
```diff
@@ -0,0 +1,53 @@
+// Get the max height and apply
+// to all textareas when page loads/reloads
+
+$(document).ready(function() {
+  var numberArray = [];
+  $('textarea').each(function(textarea) {
```
ghostPepperWonton: Can we use a class name on this? Currently this will happen on every textarea on the site, even on pages where this might not be necessary in the future. I recommend adding a new class shared between all the textareas you want to have with this behavior.

You'll probably also want to remove the textarea in the parentheses for your anonymous function. It's an unused variable and might conflict with other textareas that are named in this space.

app/assets/javascripts/autocise_textarea.js
```diff
@@ -0,0 +1,53 @@
...
+    $(this).height( $(this)[0].scrollHeight );
+    numberArray.push($(this).innerHeight());
+    var max = Math.max(...numberArray);
+    $(this).innerHeight(max);
```
ghostPepperWonton: It may not need to change, but I'm not sure I understand this. You set `height` on line 7 and `innerHeight` on this line (10). Why is that?

app/assets/javascripts/autocise_textarea.js
```diff
@@ -0,0 +1,53 @@
...
+  });
+});
+
+// Use an event to listen to all input events and then
+// filter out ones that are not on a textarea
+// if the element is a textarea we will call a
+// autoExpandTextarea function and pass in
+// the element as an argument using event.target.
+document.addEventListener('input', function(event) {
+  if (event.target.tagName.toLowerCase() !== 'textarea') return;
```
ghostPepperWonton: If you only want this behavior on a textarea input, then I would attach the event listener _only_ on that input. Additionally, since we're using jQuery on this site, it's best to create event handlers consistently.

The equivalent would look like:
```
$('textarea').change(function(event) {...});
```

app/assets/javascripts/autocise_textarea.js
```diff
@@ -0,0 +1,53 @@
...
+  autoExpandTextarea(event.target);
+}, false);
+
+var autoExpandTextarea = function(field) {
+  // check the field parameter class so we can give the
+  // same height to other textareas with same class.
+  var isMain = field.classList.contains('item-name');
+  var isMaori = field.classList.contains('item_maori_name');
```
ghostPepperWonton: Use jQuery `.hasClass()` for consistency.

app/assets/javascripts/autocise_textarea.js
```diff
@@ -0,0 +1,53 @@
...
+
+  // Reset field height
+  field.style.height = 'inherit';
```
ghostPepperWonton: You can use `.css()` in jQuery. Here's what that looks like:
```
field.css({height: 'inherit'});
```

ghostPepperWonton: This also looks like you're setting a default with this line, which should _probably_ be done in CSS. Then you can use JS to override the rule. Rather than overriding it twice.

app/assets/javascripts/autocise_textarea.js
```diff
@@ -0,0 +1,53 @@
...
+  var computed = window.getComputedStyle(field);
+
+  // Calculate the height
+  var height = parseInt(computed.getPropertyValue('border-top-width'), 10)
+               + parseInt(computed.getPropertyValue('padding-top'), 10)
+               + field.scrollHeight
+               + parseInt(computed.getPropertyValue('padding-bottom'), 10)
+               + parseInt(computed.getPropertyValue('border-bottom-width'), 10);
+  // Reset height of element using style property or just use the ifs?
+  field.style.height = height + 'px';
```
ghostPepperWonton: Same as line 31

app/assets/javascripts/autocise_textarea.js
```diff
@@ -0,0 +1,53 @@
...
+
+  // alter textarea row height to make consistent and even.
+  if(isMain || isMaori) {
+    for(var i = 0, j = 0; i < isMain.length, j < isMaori.length; i++, j++) {
+       isMain[i].style.height = height + 'px';
+       isMaori[j].style.height = height + 'px';
```
ghostPepperWonton: Same as line 31

app/assets/javascripts/autocise_textarea.js
```diff
@@ -0,0 +1,53 @@
...
+               + parseInt(computed.getPropertyValue('padding-top'), 10)
+               + field.scrollHeight
+               + parseInt(computed.getPropertyValue('padding-bottom'), 10)
+               + parseInt(computed.getPropertyValue('border-bottom-width'), 10);
+  // Reset height of element using style property or just use the ifs?
+  field.style.height = height + 'px';
+
+  // alter textarea row height to make consistent and even.
+  if(isMain || isMaori) {
+    for(var i = 0, j = 0; i < isMain.length, j < isMaori.length; i++, j++) {
```
ghostPepperWonton: If you're using the same value for `i` and `j`, I would recommend just using one. It's also best practice to give a more descriptive name to your iterators, maybe `sign`?

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -5,9 +5,9 @@
       = link_to sign_url(vocab_sheet_item.sign_id), { class: "drawing vocab_sheet_drawing drawing vocab_sheet_drawing__size--#{size}" } do
         = image_tag "signs/#{vocab_sheet_image_width}-#{vocab_sheet_image_height}/#{vocab_sheet_item.drawing}"
       = form_tag vocab_sheet_item_path(vocab_sheet_item), method: :put do
-        = text_area_tag "item[name]", vocab_sheet_item.name, class: 'item-name vocab-sheet__text-input h3', id: "item_#{vocab_sheet_item.id}_name"
+        = text_area_tag "item[name]", vocab_sheet_item.name, class: 'item-name vocab-sheet__text-input h3', id: "item_#{vocab_sheet_item.id}_name", maxLength: "100"
         = hidden_field_tag "item[#{vocab_sheet_item.id}][old_name]", vocab_sheet_item.name, class: 'old_name'
-        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item_maori_name vocab-sheet__text-input h3 maori_gloss', id: "item_#{vocab_sheet_item.id}_maori_name"
+        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item_maori_name vocab-sheet__text-input h3 maori_gloss', id: "item_#{vocab_sheet_item.id}_maori_name", maxLength: "100"
```
ghostPepperWonton: Just a small thought: Maybe this is best done as a global variable or something set by the `Item` controller?
🐘

ghostPepperWonton (commented):
Lots of little things. So close!

app/assets/javascripts/autocise_textarea.js
louise-r-blue: Had to use .Height to calculate/get the actual height and innerHeight to set the height.

app/assets/javascripts/autocise_textarea.js
louise-r-blue: for some reason would not except a class name in this instance (believe me I tried)

app/assets/javascripts/autocise_textarea.js
louise-r-blue: Its javascript

app/views/vocab_sheets/_vocab_sheet_item.html.haml
louise-r-blue: what is best done as a global variable?

app/assets/javascripts/autocise_textarea.js
louise-r-blue: its javascript - much better performance than jquery

app/assets/javascripts/autocise_textarea.js
louise-r-blue: its javascript

app/assets/javascripts/autocise_textarea.js
louise-r-blue: it s javascript

app/assets/javascripts/autocise_textarea.js
louise-r-blue: its javascript

app/assets/javascripts/autocise_textarea.js
louise-r-blue: What are the other textareas on the site?

app/assets/javascripts/autocise_textarea.js
louise-r-blue: There is also javascript on the site

app/assets/javascripts/autocise_textarea.js
louise-r-blue: mmmm

app/views/vocab_sheets/_vocab_sheet_item.html.haml
Rabid-Dan: I think he's talking about the value for max length.
I agree but instead of doing it in the controller or as a global it could just be a small wrapper method on the item model itself given that we already have access to it in the form of "vocab_sheet_item" and it should probably be a responsibility of that model anyway.
That way you could call something like vocab_sheet_item.max_length_for_name

app/assets/javascripts/autocise_textarea.js
louise-r-blue: yes right needs refactoring. The I and j act as indexes.

app/views/vocab_sheets/_vocab_sheet_item.html.haml
louise-r-blue: its a element attribute right?

app/assets/javascripts/autocise_textarea.js
ghostPepperWonton: There must be a way to get the class working properly. Have you tried using the debugger?

Other textareas include hidden fields, the notes view field, and there might be others that I'm not aware of. Using a class ensures that it's only the ones you want.

app/assets/javascripts/autocise_textarea.js
ghostPepperWonton: When you put an argument in the `.height()` method, it sets the element's height. [See docs](http://api.jquery.com/height/#height2). If you're calculating the height, doesn't that need to be assigned to a variable?

app/assets/javascripts/autocise_textarea.js
louise-r-blue: I think also this will perform better on IE

ghostPepperWonton: Is that performance measurable? I think having consistency throughout the application is more important, personally. It helps us all come to the work from the same place rather than having to rework the paradigm on each feature. This app is also suffering from being fairly staggered in terms of updates too, so the more we can bring consistency in, that will help us with feature updates.

app/assets/javascripts/application.js
```diff
@@ -16,7 +16,9 @@
 //= require search_result_link
 //= require custom_play_button
 //= require video
+//= require autocise_textarea
```
Br3nda: is that meant to be "autosize" ?

app/assets/javascripts/autocise_textarea.js
Br3nda: can you use `$('.vocab_sheet textarea')` to find it? This selector is used in vocab_sheet.js

Br3nda:
I think the textarea stuff is just for vocab sheets, yes? Can it go into `vocab_sheet.js` ? 
There's other javascript already in there modifying text area inside vocab sheet - if we put them in the same place we won't confuse future devs who can't see what other code is messing with their elements.

Br3nda:
is "autocised" a tech css term i don't know? or a mispelling?

app/views/vocab_sheets/_vocab_sheet_item.html.haml
Br3nda: Instead of putting in the controller, can it go in `VocabSheetHelper`.
```
def max_field_length
  100
end
```

Br3nda: None of it is enforced on the model though....

app/assets/javascripts/application.js
louise-r-blue: Yes thank you!. Actually should probably be autoExpand..

louise-r-blue:
Yes I totally agree about putting the textarea stuff in vocab sheets. I think I wanted to initially sit it on its own for a code review but will move

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -5,9 +5,9 @@
       = link_to sign_url(vocab_sheet_item.sign_id), { class: "drawing vocab_sheet_drawing drawing vocab_sheet_drawing__size--#{size}" } do
         = image_tag "signs/#{vocab_sheet_image_width}-#{vocab_sheet_image_height}/#{vocab_sheet_item.drawing}"
       = form_tag vocab_sheet_item_path(vocab_sheet_item), method: :put do
-        = text_area_tag "item[name]", vocab_sheet_item.name, class: 'item-name vocab-sheet__text-input h3', id: "item_#{vocab_sheet_item.id}_name", maxLength: "100"
+        = text_area_tag "item[name]", vocab_sheet_item.name, class: 'item-name vocab-sheet__text-input h3', id: "item_#{vocab_sheet_item.id}_name", max_field_length
         = hidden_field_tag "item[#{vocab_sheet_item.id}][old_name]", vocab_sheet_item.name, class: 'old_name'
-        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item_maori_name vocab-sheet__text-input h3 maori_gloss', id: "item_#{vocab_sheet_item.id}_maori_name", maxLength: "100"
+        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item_maori_name vocab-sheet__text-input h3 maori_gloss', id: "item_#{vocab_sheet_item.id}_maori_name", max_field_length
```
Br3nda: Does this work? It look like it's missing the named variable/hash key

app/helpers/vocab_sheet_helper.rb
```diff
@@ -40,7 +40,7 @@ def vocab_sheet_pages
     @vocab_sheet_pages ||= @sheet.blank? || @sheet.items.length.zero? ? 0 : (@sheet.items.length.to_f / vocab_sheet_offset_multiple).ceil # rubocop:disable Metrics/LineLength
   end
 
-  def max_field_length
+  def vocab_sheet_max_field_length
```
Br3nda: :sparkles:

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -5,9 +5,9 @@
...
         = image_tag "signs/#{vocab_sheet_image_width}-#{vocab_sheet_image_height}/#{vocab_sheet_item.drawing}"
       = form_tag vocab_sheet_item_path(vocab_sheet_item), method: :put do
-        = text_area_tag "item[name]", vocab_sheet_item.name, class: 'item-name vocab-sheet__text-input h3', id: "item_#{vocab_sheet_item.id}_name", max_field_length
+        = text_area_tag "item[name]", vocab_sheet_item.name, class: 'item-name vocab-sheet__text-input h3', id: "item_#{vocab_sheet_item.id}_name", maxLength:vocab_sheet_max_field_length
         = hidden_field_tag "item[#{vocab_sheet_item.id}][old_name]", vocab_sheet_item.name, class: 'old_name'
-        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item_maori_name vocab-sheet__text-input h3 maori_gloss', id: "item_#{vocab_sheet_item.id}_maori_name", max_field_length
+        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item_maori_name vocab-sheet__text-input h3 maori_gloss', id: "item_#{vocab_sheet_item.id}_maori_name", maxLength:vocab_sheet_max_field_length
```
Br3nda: does that maxLength work? It doesn't seem to be mentioned in the docs . It's on `text_field_tag` but not on `text_area_tag`.

https://api.rubyonrails.org/v5.1/classes/ActionView/Helpers/FormTagHelper.html#method-i-text_area_tag

app/assets/javascripts/application.js
```diff
@@ -16,7 +16,9 @@
 //= require search_result_link
 //= require custom_play_button
 //= require video
+//= require auto_expand_textarea
```
ghostPepperWonton, line 19: You're right, I like this name so much better :)

app/assets/stylesheets/_typography.scss
```diff
@@ -59,7 +59,11 @@ pre {
   padding: 15px;
 }
 
-textarea { overflow: auto; }
+textarea {
+  min-height: 1em; // ensures at least one few lines of text show up to start
+  // max-height: 8em;
+  width: 100%; //text area fills up the full width of the content area
```
ghostPepperWonton: Can we actually pull out these comments? A good rule of thumb is that the code should be able to speak for itself, and I think this does :)
🐘

app/assets/stylesheets/_typography.scss
```diff
@@ -184,6 +188,10 @@ h2+h3 { margin-top:1em; }
   display: block;
   font-size: $secondary-gloss;
 }
+.glosses-container.maori_gloss,
+.item_maori_name.maori_gloss {
+  font-size: $main-gloss;
+}
```
ghostPepperWonton: Is it possible to make these BEM?

ghostPepperWonton (commented):
I'll keep an eye out for what other reviewers think of my comment, but this should be pretty much ready to go. Just some small stuff.

app/assets/stylesheets/_typography.scss
Rabid-Dan: I think comments can have value in providing extra context for certain complex bits of code, but I agree with you here that these ones aren't really necessary. Also If the max height isn't needed anymore then can we delete it please rather than commenting it out?

app/assets/stylesheets/_typography.scss
Rabid-Dan: this will need to change once staging gets updated shortly anyway because the item_maori_name class changes to item-maori-name

Br3nda:
Looking real good!

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -5,9 +5,9 @@
       = link_to sign_url(vocab_sheet_item.sign_id), { class: "drawing vocab_sheet_drawing drawing vocab_sheet_drawing__size--#{size}" } do
         = image_tag "signs/#{vocab_sheet_image_width}-#{vocab_sheet_image_height}/#{vocab_sheet_item.drawing}"
       = form_tag vocab_sheet_item_path(vocab_sheet_item), method: :put do
-        = text_area_tag "item[name]", vocab_sheet_item.name, class: 'item-name vocab-sheet__text-input h3', id: "item_#{vocab_sheet_item.id}_name"
+        = text_area_tag "item[name]", vocab_sheet_item.name, class: 'item-name vocab-sheet__text-input h3', id: "item_#{vocab_sheet_item.id}_name", maxLength:vocab_sheet_max_field_length
```
Br3nda: does that maxLength work? It doesn't seem to be mentioned in the docs . It's on `text_field_tag` but not on `text_area_tag`.

https://api.rubyonrails.org/v5.1/classes/ActionView/Helpers/FormTagHelper.html#method-i-text_area_tag

Br3nda:
Looing a bit wonk in firefox
image

app/views/vocab_sheets/_vocab_sheet_item.html.haml
Br3nda: The max length isn't doing anything afaict. It doesn't limit in firefox or chrome. I can't see that 100 turning up in the mark up either.

Br3nda (changes requested):
maxlength isn't working.

louise-r-blue:
Yes nothing is working since changed to jquery so I'm reverting back to the javascript locally at least for now.

Br3nda:
@louise-r-blue did you see my comment? You're passing a named param to text_area_tag that doesn't exist in text_area_tag. it's in the ruby, not the js/jquery

app/views/vocab_sheets/_vocab_sheet_item.html.haml
Br3nda: https://stackoverflow.com/questions/886352/maximum-length-on-a-textarea-in-ruby-on-rails

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -5,9 +5,9 @@
...
         = hidden_field_tag "item[#{vocab_sheet_item.id}][old_name]", vocab_sheet_item.name, class: 'old_name'
-        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item-maori-name vocab-sheet__text-input h3 maori_gloss', id: "item_#{vocab_sheet_item.id}_maori_name"
+        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item-maori-name vocab-sheet__text-input h3 maori-gloss', id: "item_#{vocab_sheet_item.id}_maori_name", maxLength:vocab_sheet_max_field_length
```
Br3nda: can you make `maxLength` lowercase? This gets it working when i tested in firefox.

Br3nda (approved):

app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -173,33 +175,21 @@
     border: 1px solid $light-med-grey;
     margin-right: 11px;
     background: white;
-  
-    .item-name {
```
Rabid-Dan, line 177: Why did this and then item-maori-name classes get removed? I can see theyre still attached to their inputs in the view file so are their styles just being set somewhere else now?

Rabid-Dan:
This is a huuuugggeee 🐘 so feel free to completely disregard it, but I think it would be nice to remove the scrollbar from the text inputs now that they auto resize because it doesn't really serve any functional purpose anymore.

Rabid-Dan (approved):

app/assets/javascripts/auto_expand_textarea.js
```diff
@@ -0,0 +1,25 @@
...
+
+    $('.vocab-sheet__text-input').keyup(function() {
+      autoExpandTextarea($(this));
+    });
+
+    function autoExpandTextarea(field) {
+      var nameBeingChanged = '';
+      var height = field.prop('scrollHeight');
+
+      if (field.hasClass('item-name')) {
```
Rabid-Dan: These could be single line ifs 🐘

app/assets/javascripts/auto_expand_textarea.js
```diff
@@ -0,0 +1,25 @@
...
+      autoExpandTextarea($(this));
+    });
+
+    $('.vocab-sheet__text-input').keyup(function() {
+      autoExpandTextarea($(this));
+    });
+
+    function autoExpandTextarea(field) {
+      var nameBeingChanged = '';
+      var height = field.prop('scrollHeight');
```
Rabid-Dan: might be a good idea to rename this to something like scrollHeight so that its a little bit more obvious what it is when it gets set below 🐘

app/assets/javascripts/auto_expand_textarea.js
ghostPepperWonton: It's actually too long for CC that way :/

app/assets/stylesheets/_vocab_sheet.scss
ghostPepperWonton: The only styles in those two rules were duplicates that are now inside `.vocab-sheet__text-input` on line 153 above, so there wasn't a need to repeat the rules.

PR #280, opened 2018-08-28
BEM-ed all the classnames, restyled according to the design, made the…

ghostPepperWonton:
- BEM-ed all the classnames
- Restyled according to the design
- Made the alerts with the new style and added a transition
- Stickied the container with JS (since sticky positioning isn't yet supported in IE)
- Added the fade at the bottom of the scroll
- Made sure it doesn't flow under the footer

At the top:
screen shot 2018-08-28 at 11 59 59 am

During scroll:
screen shot 2018-08-28 at 11 59 02 am

At the bottom:
screen shot 2018-08-28 at 12 00 30 pm

New flashes:
screen shot 2018-08-28 at 11 59 26 am


louise-r-blue (approved):
like the new style

app/assets/javascripts/vocab_bar.js
```diff
@@ -1,19 +1,45 @@
...
-      bar.hide();
-      $('.not_sticky_footer').removeClass('vocab_sheet_background');
-    } else if (windowWidth > 639 &&  bar.length) {
-      bar.show();
-      $('.not_sticky_footer').addClass('vocab_sheet_background');
+  if ($('.vocab-sheet__bar').length > 0) {
+    var bar = $('.vocab-sheet__bar');
+
+    function hideVocabSheetOnMobile() {
+      var windowWidth = $(window).width();
```
Rabid-Dan: can this be achieved with a css media query? Or is there some voodoo magic in here somewhere that means it needs to be done with Javascript

ghostPepperWonton: This was actually in there before, but it could be easily taken out for CSS so I'll just do that anyway.

app/assets/javascripts/vocab_bar.js
```diff
@@ -1,19 +1,29 @@
...
+    var baseMargin = 24; // 1.5rem
+    var initialOffset = bottom + baseMargin;
+    var footerTop = $('.sticky_footer').offset().top;
+
+    function positionVocabBar() {
+      var windowScroll = $(window).scrollTop();
+      var barBottom = bar.offset().top + bar.outerHeight(true) + baseMargin;
+
+      if (windowScroll < bottom) {
+        bar.css({top: initialOffset - offset});
```
Rabid-Dan, line 16: Can this same thing be achieved with sticky/fixed css positioning? Or am I mis reading what it does?

app/assets/stylesheets/_variables.scss
```diff
@@ -46,9 +46,9 @@ $search-result-banner: 1.8em;
 }
 
 @mixin border-radius-small {
-  border-radius: 5px;
+  border-radius: .25rem;
 }
 
 @mixin border-radius-large {
-  border-radius: 25px;
```
Rabid-Dan, line 53: So kind of nit picky but I dont think the conversion here is one to one. Its probably less noticeable in the change above but there is a decent difference. I do like keeping our units rounded though. Is there a reason that these needed to change or can we just keep them as px values for now?

Br3nda (approved):
working great in Firefox.

app/assets/javascripts/vocab_bar.js
ghostPepperWonton: The short answer is yes, this is essentially sticky positioning, however we unfortunately can't use sticky because of the IE support being nil: https://caniuse.com/#feat=css-sticky

I'll look into your suggestion for using `fixed` though.

app/assets/stylesheets/_variables.scss
ghostPepperWonton: The reasoning was that I wanted to use rounded rem units so the browser does the calc. Since it was one pixel off, I wanted to use the mixins so it updates it site-wide for consistency, but I'm fine with switching it back.

Rabid-Dan: hmm I think thats some good reasoning and I'm sort of in favour of keeping the change now.
The only other issue I have with it though is that this file will now contain a mixture of em px and rem units which might be a little bit confusing from a standards point of view.

app/assets/javascripts/vocab_bar.js
Rabid-Dan: After a quick offline discussion we're also going to investigate using the foundation sticky plugin which seems to do what we want and is included in this project anyway. If that fails or there are browser compatibility issues then we'll do some minor optimisations to this and then its good

app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -250,6 +193,97 @@
...
+      position: absolute;
+      z-index: 2;
+      bottom: 0;
+      left: 0;
+      width: 100%;
+      height: 3rem;
+      background: linear-gradient(transparent, $vocab-bg);
+    }
+
+    &-header {
```
Rabid-Dan: would it be a lot of work to convert this and the other ones below from "-header" to element naming "__header"?

app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -250,6 +193,97 @@
...
+    .flash {
+      @include border-radius-small;
+    
+      margin: 0 auto;
+      background-color: $success-green;
+      width: 80%;
+      padding: .25rem;
+      opacity: 0;
+      max-height: 0;
+      transition: {
```
Rabid-Dan: I've seen a few transitions around in the last couple of PRs is there much consistency in their settings? If so it might be time to start looking at making a little mixin for the properties.

app/assets/stylesheets/_vocab_sheet.scss
Rabid-Dan: oh wait this is already nested inside an element. ignore me

app/assets/stylesheets/_vocab_sheet.scss
ghostPepperWonton: Yeah, they're all the same. Mixin time!

app/assets/javascripts/vocab_bar.js
Rabid-Dan: An additional follow up to this. We tried implementing the foundation sticky plugin but theres something in the existing styling thats preventing it from working. We've opted to backlog that particular implementation for now and get this over the line.

app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -193,93 +193,6 @@
       z-index: 11;
     }
   }
-
```
Rabid-Dan: ⭐️ 🔴 ⭐️ 🔴 ⭐️

Rabid-Dan (approved):
Great work! Thanks for working through and discussing all of my comments!

PR #283, opened 2018-08-29
Bugfix/changing names

ghostPepperWonton:
This fixes an issue where we weren't able to update English names or Maori names. Then during fixing, I found a few other issues with the way that these fields get updated, so there are actually a few files that were updated.

Note: There are also changes to the notes field because the three should really use the same implementation for updating the item.


app/assets/javascripts/vocab_sheet.js
```diff
@@ -91,51 +72,97 @@ $(document).ready(function() {
...
+      if (!!params.signId) {
+        $.ajax({
+          url: action || '/vocab_sheet/items/' + params.signId,
+          method: 'PUT',
+          data: assignItemParams(params),
+          headers: {
+            'X-CSRF-Token': $('meta[name="authenticity-token"]').attr('content'),
+          },
+        }).done(function(data) {
+          // console.log('success!', data);
```
louise-r-blue: Maybe take the console log out?

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -7,14 +7,14 @@
...
-        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item_maori_name vocab-sheet__text-input h3 maori_gloss', id: "item_#{vocab_sheet_item.id}_maori_name"
+        = text_area_tag "item[maori_name]", vocab_sheet_item.maori_name, class: 'item-maori-name vocab-sheet__text-input h3 maori_gloss', id: "item_#{vocab_sheet_item.id}_maori_name"
         = hidden_field_tag "item[#{vocab_sheet_item.id}][old_maori_name]", vocab_sheet_item.maori_name, class: 'old_maori_name'
         = submit_button('vocab_sheet.change')
     - if size == 6
       .notes-view__notes
         = form_tag vocab_sheet_item_path(vocab_sheet_item), method: :put do
           .input-with-character-count
-            = text_area_tag "item[notes]", vocab_sheet_item.notes,  {class: 'item_notes vocab-sheet__text-input h3 notes', id: "item_#{vocab_sheet_item.id}_notes", maxlength: 250 }
+            = text_area_tag "item[notes]", vocab_sheet_item.notes,  {class: 'item-notes vocab-sheet__text-input h3 notes', id: "item_#{vocab_sheet_item.id}_notes", maxlength: 250 }
```
louise-r-blue, line 17: Can maxLength be defined in the model - vocab_sheet.rb

louise-r-blue: Looking good

app/assets/javascripts/vocab_sheet.js
```diff
@@ -91,51 +72,97 @@ $(document).ready(function() {
...
   };
 
+  if ($('.vocab-sheet__text-input').length > 0) {
+    $('.vocab-sheet__text-input').keyup(function() {
+      var field = $(this);
+      action = getFormAction(field);
+      signId = getSignIdFromAction(action);
+      var data = {signId};
+
+      if (field.hasClass('item-name')) {
```
Rabid-Dan: I think all these cases can be single line ifs?
if (field.hasClass('item-name')) { data['name'] = field.val(); }
if (field.hasClass('item-maori-name')) { data['maoriName'] = field.val(); }
if (field.hasClass('item-notes')) { data['notes'] = field.val();}

app/assets/javascripts/vocab_sheet.js
```diff
@@ -91,51 +72,97 @@ $(document).ready(function() {
...
+        data['notes'] = field.val();
+      }
+
+      updateVocabItem(action, data);
+    });
+  }
+
+  function updateVocabItem(action, params) {
+    checkForTextTimeout(function(action, params) {
+      if (!!params.signId) {
```
Rabid-Dan: can we re write these to be null checks or something instead please? The double bang becomes suuper difficult to read after a while

app/models/item.rb
```diff
@@ -11,7 +11,7 @@ class Item < ActiveRecord::Base
     self.name = (sign.is_a?(Sign) ? sign.gloss_main : nil) if name.nil?
     self.maori_name = (sign.is_a?(Sign) ? sign.gloss_maori : nil) if maori_name.nil?
     self.sign_id = sign.id if sign_id.nil?
-    self.drawing = sign.drawing
+    self.drawing = sign.drawing if drawing.nil?
```
Rabid-Dan, line 14: is this still needed?

ghostPepperWonton: I liked the consistency, but I can revert it

Rabid-Dan (approved):
🎉

louise-r-blue (approved):

app/assets/javascripts/vocab_sheet.js
```diff
@@ -91,51 +72,91 @@ $(document).ready(function() {
...
-        },
-        headers: {
-          'X-CSRF-Token': $('meta[name="authenticity-token"]').attr('content'),
-        },
-      }).done(function(data) {
-        // console.log("success!", data)
-      }).fail(function(error) {
-        console.error(error.statusText);
-      });
+      action = getFormAction(elem);
```
Br3nda: missing `var` here.. ?

app/assets/javascripts/vocab_sheet.js
```diff
@@ -91,51 +72,91 @@ $(document).ready(function() {
...
+      signId = getSignIdFromAction(action);
```
Br3nda: and here, i think it's a global variable for realz - if it is, can it be in upper case to make it clearer.

app/views/vocab_sheets/_vocab_sheet_item.html.haml
Rabid-Dan: @louise-r-blue Theres only one instance of it at that value right now so I dont think its suuuuper necessary to abstract it out into a method just yet

Br3nda (changes requested):
couple variables

app/assets/javascripts/vocab_sheet.js
Br3nda: oh nah. it's a local variable. needs a `var`

app/assets/javascripts/vocab_sheet.js
```diff
@@ -91,51 +72,91 @@ $(document).ready(function() {
...
+    });
+  }
+
+  function assignItemParams(params) {
+    var data = {};
+
+    if (params.signId !== null) data['sign_id'] = params.signId;
+    if (params.name !== null) data['name'] = params.name;
+    if (params.maoriName !== null) data['maori_name'] = params.maoriName;
+    if (params.notes !== null) data['notes'] = params.notes;
```
Br3nda: gosh this bit is repetitive. Is there a way to DRY it? :elephant:

app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -187,7 +187,7 @@
       right: -14px;
     }
   
-    .item_maori_name {
+    .item-maori-name {
```
Br3nda, line 198: :sparkles:

app/assets/javascripts/vocab_sheet.js
ghostPepperWonton: You're right, getting that fixed up now

app/assets/javascripts/vocab_sheet.js
ghostPepperWonton: It is, right? The reason I did it this way was two-fold:

- The property names are different between Ruby and JS
- If the item isn't present in params, it shouldn't be in data otherwise it overrides it in Ruby (otherwise I would do a fallback assignment)

What do you think?

Br3nda (approved):

PR #292, opened 2018-08-30
Feature/alternate image only placeholder

ghostPepperWonton:
Changed out placeholder to _only_ obscure image per Micky's test


Rabid-Dan (approved):
Its a bummer that we need to remove the rest of the nice placeholder graphics but this is a good compromise between perceived speed and aesthetic.

Br3nda (approved):

PR #298, opened 2018-09-02
Vocab container height

ghostPepperWonton:
- Added an event to check that the height isn't more than the search results container.
- Made the scrollbox height of the vocab items inside the sidebar dynamic to the container height.


app/assets/stylesheets/_vocab-sidebar.scss
```diff
@@ -13,7 +13,7 @@
   z-index: 100;
 
   &:after {
-    background: linear-gradient(transparent, $vocab-bg);
+    background: linear-gradient(transparent, $vocab-bg 2.5rem, $vocab-bg);
```
Rabid-Dan, line 16: Is this line missing a comma? Or does the variable and the measurement combine somehow here?

ghostPepperWonton: After this, the height is set to `3rem`, so this measurement makes sure there is a guaranteed `.5rem` of completely blanked out background color. It was hard to see the scroll thing at smaller heights without this. Just a small tweak to make it more obvious. The padding at the bottom is a full rem anyway, so this doesn't get seen once you scroll at the way to the bottom still.

app/assets/javascripts/vocab_bar.js
```diff
@@ -5,12 +5,12 @@ $(document).ready(function() {
     var offset = searchResultBanner.offset().top;
     var bottom = offset + searchResultBanner.outerHeight(true);
     var baseMargin = 24; // 1.5rem
+    var barBottom = bar.offset().top + bar.outerHeight(true) + baseMargin;
     var initialOffset = bottom + baseMargin;
     var footerTop = $('.sticky_footer').offset().top;
 
     function positionVocabBar() {
       var windowScroll = $(window).scrollTop();
-      var barBottom = bar.offset().top + bar.outerHeight(true) + baseMargin;
```
Rabid-Dan, line 13: I've checked this locally and it looks like shifting this variable out of the method has caused it to allow for the sidebar to hide behind the footer now.
screen shot 2018-09-03 at 10 46 03 am

ghostPepperWonton: Gaaah, okay.

Rabid-Dan (approved):

louise-r-blue (approved):
looking good

PR #301, opened 2018-09-04
Feature/comment out placeholder

ghostPepperWonton:
The Jira story for this work has been rejected with this implementation as of last sprint ending on 31/08/2018. We will rework this, but currently we're disabling it for a production deploy that doesn't lose our progress.


Br3nda (approved):
There are probably better ways to do this (like in a branch). But this is temporary so let's go forwards

Rabid-Dan (approved):

louise-r-blue (approved):

PR #302, opened 2018-09-05
Remove old IE6 support file

Br3nda:
Back when Internet Explorer was a thing, it didn't support pngs.
But humanity has moved on now, so should let this file go.

http://www.dillerdesign.com/experiment/DD_belatedPNG/


Rabid-Dan (approved):
🌟

PR #312, opened 2018-09-11
Feature/remove video controls

louise-r-blue:
The user story requires that the video controls be removed so the user can view the video without any obfuscation. 
On the user clicking the start button, the button disappears, then the video starts immediately.


ghostPepperWonton (approved):

Rabid-Dan (approved):
📹 ⭐️

PR #313, opened 2018-09-12
Feature/remove video controls

louise-r-blue:
This PR includes the following.

The visibility style property has been removed and replaced with the opacity style property  as this enables the user to play and pause the video when clicking the video when accessing nzsl-online in Safari.

Added Videojs BigPlayButton (component of videos) styles specific for Chrome, Firefox and Safari means this button is removed in favour of a smaller custom play/pause icon. IE was tested in Browserstack and the intended result is similar and functional.

The auto loop has been removed from the Usage video as some of the videos are reasonably long.

Adds the videojs video controls bar back apart from a few functionalities including the feature to enable full screen mode.


app/views/signs/show.html.haml
```diff
@@ -21,7 +21,7 @@
         = add_vocab_button('vocab_sheet.item.add', class: 'add_to_sheet')
     .videos.small-12.medium-5
       .video-container
-        = videojs_rails sources: { mp4: @sign.video }, class: "main_video video normal", controlsList:"nodownload", controls: true, preload: "auto", loop: true
+        = videojs_rails sources: { mp4: @sign.video }, class: "main_video video normal", controlsList: "nodownload", controls: true, preload: "auto", loop: true
```
Rabid-Dan: These still show the play button now with the controls added back on. To meet the story requirements we'll need to remove that but leave the other controls on.

app/assets/stylesheets/_buttons.scss
```diff
@@ -69,3 +69,10 @@ a .button,
 
 .orange_submit_button { padding: 5px; }
 .translation_button { @extend .play_video_button; }
+
+.video-js.vjs-big-play-button {
```
Rabid-Dan, line 73: 🐘 but it might be worth adding a comment above these about _why_ we've chosen to hide the default play button.

Rabid-Dan (approved):

app/assets/javascripts/custom_play_button.js
```diff
@@ -25,12 +25,12 @@ $(document).ready(function() {
 
   function playVideo(video) {
     pauseOtherVideos(video);
-    $(video).closest('.video-container').children('.play-button').css('visibility', 'hidden');
+    $(video).closest('.video-container').children('.play-button').css('opacity', '0');
```
ghostPepperWonton, line 28: I have an 🐘 and a more serious concern:

🐘 : Because you're changing object properties in JavaScript, the preferred method is to use object notation, so it might be better to have it look like this:

```
$(video).closest('.video-container').children('.play-button').css({ opacity: 0 });
```

Now the real thing. By reducing the opacity, the player is still on the page. It's theoretically possible for a user to interact with this menu by clicking on it, even if it doesn't _seem_ to be there. It's probably better to use a set of properties like this and then change the class instead of assigning the CSS:

```
.play-button {
  &.hide-while-playing {
    display: none;
    pointer-events: none;
  }
}
```

Rabid-Dan: thats intended behaviour. This line only hides the play button, but we need it to remain there because when it gets clicked again the video should pause. The old method of visibility hidden worked fine in most browsers but safari had issues with it so the opacity change is a more browser cross compatible solution

ghostPepperWonton: 👍

ghostPepperWonton (approved):
Glad you were able to get that worked out. This looks so much easier than I was thinking it was going to be!

app/assets/javascripts/custom_play_button.js
louise-r-blue: I like what you are saying however this implementation means the custom icon button does not get removed(hidden) at all which is the main requirement.

PR #314, opened 2018-09-12
Rails 5 upgrade

Rabid-Dan:
This branch updates rails from 4.2.6 to ~~5.0.X (5.0.7 at this point)~~ 5.2.X (5.2.1 at this point).

This includes a bunch of syntactical changes and a whole lot of change around to config stuff.

~~I'm targeting 5.2.1 but seeing as this is a large update I'd like to get this testing against staging first before we try to make the next leap.~~


app/models/application_record.rb
```diff
@@ -0,0 +1,3 @@
+class ApplicationRecord < ActiveRecord::Base
+  self.abstract_class = true
+end
```
ghostPepperWonton, line 3: Is there a `.editorconfig` file for this repo? There's no final newline here.

app/views/layouts/mailer.html.haml
```diff
@@ -0,0 +1,6 @@
+!!!
+%html
+  %head{:lang => "en"}
```
ghostPepperWonton: Correct me if I'm wrong, but I thought in HTML5, the `lang` attribute goes on the HTML tag, right?

KatieWright26 (approved):

spec/controllers/sign_image_controller_spec.rb
```diff
@@ -7,25 +7,25 @@
 
   describe '#show' do
     context 'good request' do
-      before { get :show, valid_params }
+      before { get :show, params: valid_params  }
```
ghostPepperWonton, line 10: I see similar things below. Is this different then `params[:valid_params]`?

Rabid-Dan: valid params is set above and output within curly braces already so this is the equivalent of params: { id: 2, name: "bob" }

Rabid-Dan:
Just as a quick side note on this Vini is going to pull the branch down and run a manual check to see if anything looks off or runs unexpectedly. She has a great ability to catch random UI things that are often overlooked so I'll hold off merging this until she's given it a look over.

lamorrison (approved):

app/views/layouts/mailer.html.haml
Rabid-Dan: I think you're right, its set on the head in the application layout but thats also possibly and error so I'll change em

ghostPepperWonton (approved):

Br3nda:
FABULOUS WORK!

Gemfile
```diff
@@ -3,7 +3,7 @@
 source 'https://rubygems.org'
 ruby '2.5.1'
 
-gem 'rails', '~> 4.2.6'
+gem 'rails', '~> 5.0.0'
```
Br3nda: Do you want to take this higher before merging? 5.0.0 has some known vulns

app/views/layouts/application.html.haml
```diff
@@ -1,6 +1,6 @@
 !!!
-%html
-  %head{:lang => "en"}
+%html{:lang => "en"}
```
Br3nda: :sparkle:

Br3nda: can use the new ruby hash syntax too?

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: redis://localhost:6379/1
```
Br3nda: do we have redis?

Br3nda (dismissed):
looks amazing.. needs to be a higher rails version, so we don't have the known vulns in rails 5.0 going to production

Br3nda:
(dismiss my review when you reckon you've got a secure version of the rails gem)

config/cable.yml
Rabid-Dan: hmm I dont think we do. This was one of the files that got added by the app:update process so I can go back and add redis in now if you think its worth it?

Rabid-Dan:
@Br3nda Thanks! the jump from 4.2 to 5 is pretty big so what do you think about doing a staggered rollout to staging with 5.0 first then 5.2 after incase theres any weirdness with how it operates in heroku? I'm working on a 5.2 update now thats branched off this so alternatively I can just merge that into this when its ready?

Br3nda:
@Rabid-Dan So you want to get it as far as staging? You could deploy to a new heroku temporarily. The app should deploy pretty much out of the box that way. Deploy master, and then deploy this branch over top to see.

config/cable.yml
Br3nda: nah, just not sure why we have redis config if we're not using redis.

Br3nda:
:+1: Sweet. I'll test it tonight.

Br3nda:
actually - go ahead and merge if you have reviewers here already :-D

Rabid-Dan:
@Br3nda do you want to merge the Release15 branch into master before I merge this into staging? Otherwise that release will need to wait until we've confirmed that the upgrade is working without issue

Br3nda:
Done. Staging code is going to prod now

PR #319, opened 2018-09-17
Auto corrected by following Ruby MagicComment

Br3nda:
Auto corrected by following Ruby MagicComment

Click [here](https://awesomecode.io/projects/58/ruby_config_groups/544) to configure it on awesomecode.io


Rabid-Dan (approved):
🙌

PR #322, opened 2018-09-17
Bugfix/show back button

louise-r-blue:
I had 'why didn't I think of that moment' when @Rabid-Dan saw that the existing logic in the search page results card 'add to vocab sheet' can be reused.

This PR resolves the story that a user could not see their updated vocab sheet when adding signs and navigating back to the search results page.

The 'arrow icon', Back to search results' link and text was not rendering when a user added a sign to their vocab sheet.

This is history

I have added then removed some jquery to the existing HideBackButton fn that creates a string to append a link, text and a history api call to the back button container so the user can click and navigate back to the previous page.


louise-r-blue:
ok I have a better way of fixing this so no pr's just yet :/

vinithanatarajan (approved):

app/helpers/application_helper.rb
```diff
@@ -46,20 +46,6 @@ def play_video_button(text, url = nil, options = {})
             { class: "button #{options[:class]}" }.reverse_merge(options)
   end
 
-  def add_vocab_button(text, url = nil, options = {})
```
Rabid-Dan, line 49: ⭐️

app/assets/javascripts/search_result_link.js
```diff
@@ -4,7 +4,7 @@ $(document).ready(function() {
   });
 
   function hideBackButton() {
-    if ($('.back_to_search_results').length === 0) {
+    if ($('.back_to_search_results').length === 0 ) {
```
Rabid-Dan: can you remove the extra space at the end here please

app/assets/stylesheets/_search_result.scss
```diff
@@ -55,7 +55,11 @@
     .notes,
     .videos {
       float: left;
+      padding-top: 5px;
```
Rabid-Dan: now that the back button is never hidden does this extra padding still need to be here?

app/assets/stylesheets/_search_result.scss
```diff
@@ -55,7 +55,11 @@
...
       video { margin-bottom: 10px; }
     }
   }
 }
+.search-result-container {
```
Rabid-Dan: it looks like this doesnt actually do anything now? Is it still needed for anything?

app/assets/stylesheets/_search_results.scss
```diff
@@ -71,6 +71,11 @@
     width: 100%;
   }
 
+  &__add-to-vocab-sheet {
```
Rabid-Dan: I think this should be an element on the sign-container class rather than the card class. The card class is used almost exclusively on the search results page right now and has its own equivalent button which is positioned slightly differently, so this could get confusing in the future as an element of that.

app/assets/stylesheets/_search_result.scss
louise-r-blue: There is a conditional in the checkVocalBarHeight function in the `vocab_bar.js` file that utiltizes the class on line 23 so I should probably leave in.

```
function checkVocabBarHeight() {
      if (barBottom > footerTop) {
        if ($('.search-results__container').length > 0) {
          bar.css({ height: $('.search-results__container').innerHeight() });
        } else if ($('.search-result-container').length > 0) {
          bar.css({ height: $('.search-result-container').innerHeight() });
        }
      }...

```

Rabid-Dan: but that just means it will only need to be left in the view. You can take this style out of the css file because it doesnt serve any _styling_ purpose. The class can still stay on an element in the view without having anything attributed to it in a css file.

app/assets/stylesheets/_search_results.scss
louise-r-blue: Thanks and yes I will change. I have the .sign-container appearing in 3 different scss files _search_results.scss, _home.scss, and _search_result.scss ( this one) and appears in 4 view files including home.html.haml, _sign_of_the_day.html.haml, search.html.haml (this one) and show.html.haml

app/assets/stylesheets/_search_result.scss
louise-r-blue: it has a position style on which does not show in this pr but I can move it further up the file

louise-r-blue: Actually I can't move it to what I though was a good place is a specific media style but I CAN extract the style and place in an appropriate place.

app/assets/stylesheets/_search_result.scss
```diff
@@ -4,6 +4,9 @@
     padding: 25px 0;
   }
   &-container {
+  {
```
Rabid-Dan: I guess my original comment about the position relative attribute is that I cant see if serving any functional purpose. That is to say, adding this property doesnt actually change the page from what I can see so I dont think its actually needed.

app/views/signs/show.html.haml
```diff
@@ -16,9 +16,10 @@
       .glosses-container.glosses
         %h2.main_gloss= @sign.gloss_main
         %h2.main_gloss.maori-gloss= @sign.gloss_maori
-      = form_tag vocab_sheet_items_path do
-        = hidden_field_tag 'sign_id', @sign.id
-        = add_vocab_button('vocab_sheet.item.add', class: 'add_to_sheet')
+      = link_to "#", { class: "card__button add-to-vocab-btn ""add-to-vocab-sheet"" card__button--peach", "data-sign-id": @sign.id } do
```
Rabid-Dan: You've got some unnecessary double quotes around add-to-vocab-sheet on this line.
I think this will technically still function but both of the pairs before and after that class should be removed for readability.

app/assets/stylesheets/_buttons.scss
```diff
@@ -34,6 +34,11 @@
   }
 }
 
+.add-to-vocab-sheet {
```
Rabid-Dan: Just did some test on this locally and this whole style isnt actually needed. The card__button class overrides the float: left with its own, and the font size shouldnt be made larger based on what staging and prod look like.

app/assets/stylesheets/_buttons.scss
```diff
@@ -1,5 +1,5 @@
+.add-to-vocab-btn,
 .add_vocab_button {
```
Rabid-Dan: oh sorry one tiiinny last thing. It looks like the add_vocab_button class isn't actually used anywhere anymore since deleting that helper method, so you can get rid of the class here now and just leave these styles assigned to the add-to-vocab-btn class instead.

app/assets/stylesheets/_search_results.scss
```diff
@@ -71,7 +71,8 @@
     width: 100%;
   }
 
-  &__button {
+  &__button,
+  .sign__button {
```
Rabid-Dan: The .sign__button class is another one that looks like it isn't used anywhere? You should be able to remove that as well now and just have these styles assigned to the &__button element.

ghostPepperWonton (approved):

Rabid-Dan (approved):

PR #327, opened 2018-09-18
Add ability to download PDF copies of vocab sheets.

Rabid-Dan:
User story in Jira ticket NZSL-14 in the Ackama Jira board:

As an end user with a vocab sheet
I can save/download my vocab sheet
So that I can print/view/distribute it outside of the website

Originally I had tried to use wicked_pdf to handle PDF generation, however there were some pretty bad rendering issues that came up because of the way that images are pulled in and stored in the public directory and the implementation quickly grew in complexity to accomodate that. After fighting that for a while I decided to take a different route and bring over an implementation from another Ackama project that consists of a single PDF rendering service class and a handful of surrounding infrastructure files. They use a combination of Chrome and puppeteer to convert an html version of the page into a pdf version.
 
While its a bit of additional code, it is all fairly neatly tied up and works super well without any modification. At some point I'd like to bundle it all into a gem for better re-use as well.

The usability changes for this include the addition of a new "Download as PDF" button on the vocab sheet page, and a flash style message to alert progress to the user when that button is clicked.

The introduction of puppeteer also means the introduction of yarn. Because of this I've had to adapt the travis setup to get all the tests passing (which mostly just includes specifying node and yarn versions and the introduction of an environment variable to run puppeteer in travis).

Gif below:
 
wooooaahhh-pdf


louise-r-blue:
Looking great. Micky is going to be over the moon. One question why do we need a secrets file for this

louise-r-blue (approved):

.codeclimate.yml
```diff
@@ -6,3 +6,7 @@ engines:
         enabled: false
       require-jsdoc:
         enabled: false
+exclude_patterns:
+- "bin/render-pdf.js"
+- "db/migrate/20110919012833_create_pages.rb"
```
Rabid-Dan, line 11: I dont know if we want to run the checks across migrations for any reason so I've disabled just these two for now as they were throwing duplicated code errors.

app/controllers/vocab_sheets_controller.rb
```diff
@@ -12,8 +12,18 @@ def show
...
 
-    return render :print if params[:print] == 'true'
-    render :show
+    respond_to do |format|
+      format.html do
+        return render :print if params[:print] == 'true'
+        render :show
+      end
+
+      format.pdf do
```
ghostPepperWonton, line 18: ✨

ghostPepperWonton (approved):

louise-r-blue (approved):
is that what was stopping the CI build

.codeclimate.yml
```diff
@@ -6,3 +6,8 @@ engines:
...
+- "db/migrate/20110501210932_create_feedbacks.rb"
+- "spec/controllers/admin/pages_controller_spec.rb"
```
Rabid-Dan, line 13: I've also added the pages controller spec to this list. I'm not super happy about it but its been throwing code duplication errors for the edit action tests which are similar but deliberately so.

Rabid-Dan:
@Br3nda would you mind taking a look at this if you get a bit of time? There are a handful of reasonable additions in it so it would be good to get your opinions. If you're stretched for time then I can always get some extra reviews from people here as well.

app/services/pdf_rendering_service.rb
```diff
@@ -0,0 +1,121 @@
...
+    path = File.join(TMP_DIR_PATH, "#{@unique_id}-output.pdf")
+    FileUtils.touch(path)
+    path
+  end
+
+  def render_as_pdf(input_html_path:, output_pdf_path:)
+    cmd = "node #{Rails.root.join("bin", "render-pdf.js")} #{input_html_path} #{output_pdf_path}"
+    Rails.logger.info(cmd)
+
+    error_msg = "Chrome did not complete the PDF conversion within the #{PDF_CONVERSION_TIMEOUT_SECONDS} second timeout"
```
fractalcactus, line 83: Are you certain that the browser is always Chrome?

app/controllers/vocab_sheets_controller.rb
```diff
@@ -51,4 +58,17 @@ def set_vocab_sheet
     def set_title
       @title = @sheet.name
     end
+
+    def build_rendered_pdf(html:)
+      renderer = PdfRenderingService.new(from_html: html)
+      renderer.render
+      renderer.pdf
```
fractalcactus, line 65: 🐘 why separate out `renderer.render` and `renderer.pdf`, when `renderer.render` returns the pdf? 
`renderer.render` _could_ be renamed `renderer.get_pdf`, and line 65 could be removed. 🔬 in the scheme of things though : )

bin/render-pdf.js
```diff
@@ -0,0 +1,38 @@
...
+const puppeteerArgs = process.env["CI"] ? ["--no-sandbox"] : [];
+
+console.log("Input path:", inputPath);
+console.log("Output path:", outputPath);
+
+(async () => {
+  const browser = await puppeteer.launch({args: puppeteerArgs});
+  const page = await browser.newPage();
+
+  console.log("Starting PDF conversion");
```
fractalcactus, line 27: is this meant to be left in prod?

fractalcactus:
Wow, seems to me that you've written super clean nice code in the face of a complex problem. 'twas a pleasure to review! I left some small comments, but overall, looks amazing

Rabid-Dan:
@fractalcactus thanks! Unfortunately I can't take credit for it because its 99% Eoins implementation from SafePlus.

spec/services/pdf_rendering_service_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  describe "#render" do
+    let(:pdf_file_header_bytes) { "%PDF-1.4" } # PDFs begin with this sequence of bytes
+
+    it "adds a base tag immediately after the <head> opening tag in the given HTML" do
+      subject.render
+      expect(subject.html).to match(/<head><base href=/)
+    end
+
+    it "Saves a PDF version of the HTML file to disk" do
+      subject.render
```
Br3nda: how about moving this to a before { } block. 
then the 3 expects will be run separately - we'll get much more useful messages on that day in the future when we break something and the tests fail.

Br3nda:
```ruby
    describe "Saves a PDF version of the HTML file to disk" do
      before do
        subject.render
        pdf_file_path = subject.pdf.file_path
        first_bytes_in_file = File.read(pdf_file_path, 8)
      end
      it { expect(subject.html).to match(/<head><base href=/) }
      it { expect(File.exist?(pdf_file_path)).to eq(true) }
      it { expect(File.size(pdf_file_path)).to be > 0 }
      it { expect(first_bytes_in_file).to eq(pdf_file_header_bytes) }
    end
```

app/controllers/vocab_sheets_controller.rb
Rabid-Dan: hmm I kind of agree, although .pdf here ends up being a new instance of RenderedPdf while the render method runs everything required to get it to that point so I kind of like the separation.

app/services/pdf_rendering_service.rb
Rabid-Dan: While this does function in other browsers, it uses chrome tech for the pdf rendering and I think this message is just for the logs so it hopefully shouldnt be available to confuse users

bin/render-pdf.js
Rabid-Dan: Yeah I think mostly just for logging purposes

spec/services/pdf_rendering_service_spec.rb
Rabid-Dan: oh yeah thats a great idea!

Br3nda:
Ready to merge this?

PR #332, opened 2018-09-18
Feature/increase notes glossfield

louise-r-blue:
An update to this PR:
On comments and advice  received I have reversed the changes I made and instead made a change on one class the `notes-view__gloss` class. This change widens the gloss field so that a word with a length of 17 characters such as  `electrocardiogram` can fit on one line. 

revert_changes

This PR helps to resolve a user story NZSL-7 that required the gloss fields be increased in width so that long glosses such as `electrocardiogram` can comfortably fit on one line. 

Note: Gloss refers to the 2 smaller textfields on the left of the notes view. The top one is the Main gloss and the bottom one is the Maori gloss. The screenshots show the change in notes view and print view.

notes_print
notes_view

I kept in mind  keeping with-in confines of the container and maxing out gloss textfield areas.

- decreased the padding with-in the glosses textfields.
- decreased the padding in the outer container holding the glosses and drawing.
- decreased the max-width of the notes view ( this does not lessen the max character length ).


app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -104,15 +104,19 @@
 
         .notes-view {
           &__gloss {
-            max-width: 10rem;
```
Rabid-Dan, line 107: This is a totally valid way to approach this problem but I think I'd prefer if we kept the padding as is so it gives the UI a bit of breathing room.
You can achieve a similar solution while maintaining the existing padding by modifying the flex and max width properties (you also don't have to set a width because flex is handling that) for just the gloss class to 20rem each.

app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -104,15 +104,19 @@
...
+            .item-maori-name,
+            .item-name {
+              padding: .3rem .2rem .3rem .2rem;
```
ghostPepperWonton: 🐘 This can be simplified to `padding: .3rem .2rem;`

app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -104,15 +104,19 @@
...
-            padding: .5rem 1rem .5rem .5rem;
+            padding: .5rem .55rem .5rem .25rem;
             border-right: 1px solid $light-med-grey;
           }
 
           &__notes {
-            max-width: 90%;
-            flex: 90% 9;
+            max-width: 65%;
+            flex: 65%;
```
ghostPepperWonton: Doesn't this still need a `flex-grow` value? This was 9 before, is it 6 now?

app/assets/stylesheets/_vocab_sheet.scss
louise-r-blue: Thank you!

louise-r-blue (commented):
Thanks I am reverting this change

vinithanatarajan (approved):

Rabid-Dan (approved):
👏 👏 👏

PR #333, opened 2018-09-19
Change over images to new higher resolution format and display in all instances

Rabid-Dan:
User story from Ackama jira ticket NZSL-26

```
As a user

I want to see images that are un cropped and have original dimensions of 400 x 400

So that the image quality is higher than its current state
```

Historically images have needed to be cropped due to some offset issues. More recently a new set of clearer images, that have a max width or height of 400px and dont require cropping, have been created and uploaded to freelex for use. 

Ideally the NZSL team would like to use these as the default images for everything if possible so this branch changes all instances of images pulled from freelex to the new dimensions.

Due to the increase in size and clarity we'll need to do some benchmarking to make sure that page load times dont increase dramatically. The new images are larger but are still not huge, so load time stats will be taken against staging before this branch is merged and then again after this has been merged for comparison.

This piece of work also means that a bunch of old helper methods can be removed as they were previously relied on for calculating image dynamic image dimensions in the vocab sheet and have now been removed.


Rabid-Dan:
I'll take a second pass on this tomorrow morning and maybe add some more tests, but reviews on the initial implementation are welcome.

app/services/image_processor.rb
```diff
@@ -14,24 +14,16 @@ def initialize(filename:, height:, width:)
...
     return @local_filename if local_file_exists && local_file_age_in_days < 1
-    resize
+    write_file_locally
     @local_filename
   end
 
   private
 
-    def resize
+    def write_file_locally
```
Br3nda, line 25: noice rename

louise-r-blue: I agree

louise-r-blue (approved):
an amazing implementation and tidy up.

app/assets/stylesheets/_vocab_sheet.scss
```diff
@@ -196,19 +196,31 @@
 .vocab_sheet_drawing__size {
   &--2 {
     min-height: 290px;
```
fractalcactus: my scss is rudimentary - so 🐘 , but this repetition of 
```max height: number &img max-height: number - 10 ``` could be dry'd up with a mixin?

File: app/controllers/sign_image_controller.rb [hunk 3]
```diff
@@ -25,13 +25,13 @@ def filename_param
     def width_param
       sign_image_params[:width].to_i
     rescue StandardError
-      100
+      400
```
fractalcactus, line 28: nice!

app/assets/stylesheets/_vocab_sheet.scss
Rabid-Dan: oooh thats a great idea thanks!

fractalcactus (approved):

KatieWright26 (approved):

app/helpers/vocab_sheet_helper.rb
```diff
@@ -5,30 +5,6 @@ def vocab_sheet?
...
-  end
-
-  def vocab_sheet_label_fontsize
-    @vocab_sheet_label_fontsize ||= (vocab_sheet_label_height.to_f / 2).floor
-  end
-
-  def vocab_sheet_item_width
-    @vocab_sheet_item_width ||= vocab_sheet_image_width + 5
-  end
-
```
ghostPepperWonton, line 31: Aaaawww yeaaaaah, dropping these is so exciting

ghostPepperWonton (approved):

Br3nda (approved):

PR #348, opened 2018-09-23
configure puppeteer to launch in no sandbox mode for running on heroku

Rabid-Dan:
Puppeteer has issues launching chrome on heroku (see here: https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-on-heroku )

To work around this we need to install an extra build pack to get things running which can be viewed here https://github.com/jontewks/puppeteer-heroku-buildpack.

Part of the requirement for that buildpack is to set the no-sandbox argument when launching puppeteer so I've made that change here and left a comment in the file indicating why its needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ sudo: false
 cache: bundler
 env:
   global:
-  - CI=true  
   - S3_BUCKET_URL=http://s3-ap-southeast-2.amazonaws.com/test-asset-bucket/
   - GIT_COMMITTED_AT=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then git log -1 --pretty=format:%ct;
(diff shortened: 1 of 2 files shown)


ghostPepperWonton (approved):

louise-r-blue (approved):
good work!

Br3nda (approved):
Beautiful. Ship it.

Rabid-Dan:
Just as an update to this, the --no-sandbox option is a bit insecure so I'm asking around to see if theres something else I can do instead. Will hold off merging until I've had those discussions

Rabid-Dan:
Alright have had some good explanation about why Heroku needs this and it sounds like its probably the best/only option for getting things running for now, so ill merge it.

PR #349, opened 2018-09-24
Remove class selector to return to default html cursor behaviour

louise-r-blue:
The user story NZSL-38 requires that when the user clicks the enter button on their keyboard the cursor will move to the next line in any of the editable textareas.

An HTML default behaviour for a textarea allows the cursor to move to a newline but this had been overridden in the vocab.js file.  Removing a class selector on the keypress function in this file enabled the default behaviour.


Rabid-Dan:
This will need a description for other reviewers.

ghostPepperWonton:
Doesn't this still submit the form? You may need a separate keypress handler to prevent the default behavior still.

louise-r-blue:
Thanks @danieljauch for your comment. I did have this event.preventDefault in a keydown fn but I will take  a further look. 

@danieljauch After a conversation with @Rabid-Dan I will keep to the current implementation. Users will expect to be able to click enter and move the cursor to a new line. I don't think we can have a preventDefault in the same implementation. I have tested this across browsers that if a user hits enter when in a editable textarea their changes are persisted including if focus is outside the textarea, navigating back, reloading the site and printing. The first screenshot are changes I did yesterday. What do you think? 

chrome_newline_persists
firefox_newline_persist
safari_newline_persisted

Rabid-Dan (approved):

ghostPepperWonton (approved):

app/assets/javascripts/vocab_sheet.js
```diff
@@ -30,7 +30,7 @@ $(document).ready(function() {
         }
       };
 
-      $('.vocab_sheet textarea, input.vocab_sheet_name').keypress(function(e) {
+      $('input.vocab_sheet_name').keypress(function(e) {
```
Br3nda, line 33: :sparkler:

Br3nda (approved):

PR #351, opened 2018-09-24
Add staging environment

Rabid-Dan:
The new save as PDF work requires an app domain name and protocol which are stored against the environment in the secrets file. I think this is causing issues on staging because while settings exist for staging in a bunch of places, staging just uses the production settings instead.

This branch adds a staging env file as part of the process of changing that over. 
With how everything else is configured it looks like the only thing that should change after deployment is those secrets items.

### DIFF
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+Rails.application.configure do
+  # Settings specified here will take precedence over those in config/application.rb.
+
+  # Code is not reloaded between requests.
+  config.cache_classes = true
+
+  # Eager load code on boot. This eager loads most of Rails and
+  # your application in memory, allowing both threaded web servers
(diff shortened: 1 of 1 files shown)


ghostPepperWonton (approved):

elspeth-rabid (approved):

elspeth-rabid:
@joshmcarthur just made the valid point that heroku uses the production rails environment, always.

vinithanatarajan (approved):

Rabid-Dan:
oooooh @elspeth-rabid will this not do what I want it to do then? I could probably change the items that are stored in the secrets file to just be straight environment variables and save them in the heroku config variables instead

elspeth-rabid:
@Rabid-Dan yeah I think that is the Heroku way.

Rabid-Dan:
rad! thanks for the heads up. I'm gonna close this one and open up a new one to do that instead

PR #353, opened 2018-09-24
Allow for accessing of sign images without auth

Rabid-Dan:
PDF generation is currently blocked by images not generating properly.
This is caused by their urls requiring basic auth credentials, where other assets that have been precompiled do not (meaning styles and images served from the assets folder are fine).

I've added a skip before action to for the sign images controller show action only to move around this.
This feels like its a safe enough thing to do because image urls are pretty complicated to put together, however if there are concerns for security around this then I'm more than happy to make changes.


Br3nda:
You could put the user and pass into the url configured on staging in the env variables.

Br3nda:
Also, we can run on review apps, so you can run your pr branch on heroku, and see if it works before merging into staging. It can save a lot of back and forth.

ghostPepperWonton (approved):

Rabid-Dan:
Thanks @Br3nda thats super helpful. Josh had tried to say the same thing to me yesterday but I didn't properly take the information onboard, I'm going to close this branch because its not needed anymore.

PR #357, opened 2018-09-24
Add an explicit height to vocab sidebar images for consistency

louise-r-blue:
This helps resolve NZSL-46. 

There is actually 2 parts to this story which is really about consistent image resizing. . Firstly the  vocab-side bar image resizing and secondly the sign detail page image and gloss container resizing.

In response to further discussion with the client this morning the sign container with the drawing and glosses on the sign detail page have been resized for consistency across other sign detail drawings. I would have given each image/drawing its full width and height had it not been for the sign attributes in the top left corner. Of course a sacrifice is the extra white space above the glosses as in first screenshot.

container 459 * 602
image/drawing 325 * 400
325_x_400

container 459 * 602
image/drawing 350 * 240
350_x_240

This story required that an explicit height be set to the vocab side bar image so the user can view consistently sized drawings. I chose to use rem units as it seemed to give the most consistent height across browsers. The attached screenshot has the same result in Firefox, Safari, Chrome, Windows 8, 8.1, and 10 on IE 11 with the height of all the drawings ( I chose differently sized drawings to test) originally at 5.625rem which is about 90px but updated to 5.5rem @ 88px;

height_rem


app/assets/stylesheets/_vocab-sidebar.scss
```diff
@@ -56,6 +56,7 @@
   &__image {
     min-height: 5rem;
     max-width: 85%;
+    height: 5.625rem;
```
Rabid-Dan: Can we make this a rounder number like 5.5 or 6? 5.625 seems a little finicky unless theres a reason that it needs to be that size.

louise-r-blue: Ok sure I will make it 5.5rem. The only reason was that 5.625 matched 90px exactly

ghostPepperWonton (approved):

Rabid-Dan (approved):
Approved but just a reminder not to merge this yet.

Rabid-Dan (changes requested):
I think all of the button style changes are a bit over complicated for the original intent of the story. I'm happy to pair with you on this if you'd like, but we should be able to fulfil the story requirements through height and width changes only (unless I'm missing a particular use case, which is _highly_ possible)

louise-r-blue:
This branch has been replaced with a new branch and better implementation https://github.com/ODNZSL/nzsl-online/pull/366

PR #358, opened 2018-09-24
Allow authentication for basic auth through puppeteer

Rabid-Dan:
Images are failing to render within downloaded pdfs on staging because basic auth is preventing them from being accessed during the rendering process.

Embedded credentials are being blocked in chrome now so unforunately its no longer possible to change the base url of the images to user:pass@url.com to get around this.

Puppeteer has an authentication method that can be called though so this branch facilitates the accessing and passing of credentials from environment variables by the pdf rendering service, and then utilisation of those credentials by puppeteer if non nil values are passed.

Through replicating the conditions locally and testing with and without credentials this method appears to work.


louise-r-blue (approved):
Good work!!

bin/render-pdf.js
```diff
@@ -12,6 +12,8 @@ const puppeteer = require("puppeteer");
 
 const inputPath = `file://${process.argv[2]}`;
 const outputPath = process.argv[3];
+const basic_auth_username = process.argv[4];
+const basic_auth_password = process.argv[5];
```
ghostPepperWonton, line 16: There's no way to turn these into symbols is there?It would be nice to have something like `process[:username]`

ghostPepperWonton (approved):

bin/render-pdf.js
Rabid-Dan: I dont think so? I did try to put both of the new ones into a single credentials argument but there was extra work to convert them back into something usable from there unfortunately.

PR #364, opened 2018-09-25
Stop no index meta tag from rendering except for on staging environment

Rabid-Dan:
During the course of our current block of SEO improvement work I've found that the robots noindex tag is being rendered for all environments. 
This is definitely something that we don't want to do and is counter to our current work so here i'm re-implementing an old helper method so that we only render that tag on staging.


ghostPepperWonton (approved):

louise-r-blue (approved):
Great : )

PR #367, opened 2018-09-27
Update/download images

louise-r-blue:
This is the implementation for NZSL-44 . This user story requires that the user is able to download high resolution images/drawings from their vocab sheets and from the sign details page. 

I suspect the client may wish to alter the position of the download button on the users vocab sheets but this implementation follows the clients acceptance of the initial design.

In some cases I have used rem over px but not for units under 5px because it gets a little too finicky ( or should I be consistent  .31225rem versus 5px ).

If I have used something like 2.1 rem as opposed to 2rem I was probably trying to be too precise. As an example on the sign details page I had the download btn at 2.1rem. However when it came to equalising the height of the 2 buttons I had to revert to px for the download button (screenshot 3). The add button at 33px and the download button at  2.0625rem which I rounded to 2.1rem meant these buttons were visibly not the same height - as you may see in screenshot below.

screen shot 2018-09-27 at 1 43 23 pm
screen shot 2018-09-27 at 1 43 48 pm
equal_height_buttons


app/assets/stylesheets/_search_result.scss
```diff
@@ -37,22 +38,28 @@
...
+      margin: 0 0 10px 0;
     }
   }
 }
 
 @media screen and (min-width: $medium-screen) {
   .search-result-container {
     .sign-container { float: right; }
-    .add-to-vocab-btn,
+    .add-to-vocab-btn { float: right; }
```
ghostPepperWonton: 🐘 Should the positioning for this file change to flex? Is that on the horizon with another branch?

ghostPepperWonton (approved):

app/views/signs/show.html.haml
```diff
@@ -16,10 +16,15 @@
...
-      = link_to "#", { class: "card__button add-to-vocab-btn card__button--peach", "data-sign-id": @sign.id } do
-        %span.icon-container
-          %i.fi-plus
-        %span.text-container Add to Vocab Sheet
+      .button-container
+        = link_to "/images/signs/400-400/#{@sign.drawing}", { "data-sign-id": @sign.id } do
+          %span
+            = image_tag "/assets/ui/download-button.svg/", class: "download-button"
+          %span.button-action  Download Drawing
+        = link_to "#", { class: "add-to-vocab-btn sign__button--peach", "data-sign-id": @sign.id } do
```
Rabid-Dan: why did the classes on this get changed/removed? I cant see any styles for the sign__button--peach class that looks to have replaced the original card__button--peach class?

app/assets/stylesheets/_search_result.scss
```diff
@@ -38,22 +39,28 @@
...
-    .examples-container {
-      .translation-gloss {
-        @include montserrat-regular;
-        font-size: $h6-font-size;
-        margin: 0 0 10px 0;
-      }
+    .add-to-vocab-btn { padding-right: 45px; }
+  }
+
+  .examples-container {
```
Rabid-Dan, line 46: is there a reason this block was pulled a level up from being nested in the search-result--container class to just being nested in the search-result classe? Also can the indentation of all of this be tidied up please.

app/assets/stylesheets/application.scss
```diff
@@ -57,7 +57,6 @@ li { margin: 0; }
 
  img {
    display: inline;
-   margin: 5px 10px;
```
Rabid-Dan: This looks to be for all image tags across the site. Are you sure that removing these margins wont affect any existing images? It might be safer to set a margin 0 on an exisiting image class for the specific images you want to have that zero value.

app/assets/stylesheets/_search_result.scss
Rabid-Dan: +1 to this, Floats can be pretty finicky. Big thumbs up for using flex instead if we can

app/assets/stylesheets/_search_result.scss
```diff
@@ -38,22 +39,28 @@
           margin: 10px auto;
           img { margin: 0; }
         }
-      }
+       }
+     }
+    .button-action {
+      float: left;
```
Rabid-Dan: Is it possible to use a display:inline-block instead of a float left here? This class only looks to be set on the wording for each button? Or is there another float that set somewhere else that this is trying to override?

app/assets/stylesheets/_buttons.scss
```diff
@@ -68,6 +67,22 @@ a .button,
 
 .orange_submit_button { padding: 5px; }
 .translation_button { @extend .play_video_button; }
+.download-button,
+.vocab-download-button {
+  float: left;
```
Rabid-Dan: These are just for the images right? Is there a reason they need floats to be positioned?

app/assets/stylesheets/_buttons.scss
```diff
@@ -68,6 +67,22 @@ a .button,
...
+  margin-left: none;
+}
+
+.download-button {
+  height: 33px;
```
Rabid-Dan: Can these be converted over to rem units rather than px units please? I think we agreed a sprint or two ago to start using them as a standard unit.

Rabid-Dan (changes requested):
A few questions in there. I'd like to see us avoid floats if we can because they can be a bit finicky to work with. Maybe not possible depending on what existing styles are doing but worth investigation anyway.

app/assets/stylesheets/_search_result.scss
louise-r-blue: @Dan @danieljauch I have added an existing class with a position of flex. Putting flex on this `.add-to-vocab-btn` resulted in weird positioning.

app/assets/stylesheets/_buttons.scss
louise-r-blue: @Dan I did explain in my PR above. The existing 'Learn more' btn has 33px and if I was to be use rems it would be 2.0625rem which I questioned as being too finicky. Rounding to 2.1 did not mean the buttons were the same height unless I have the go ahead to use 2.0625rems

app/assets/stylesheets/_search_result.scss
louise-r-blue: @Dan Code Climate did not like the nesting. Only did not complain when I pulled it out of the search results class.

Rabid-Dan: Ok it looks like its only used in one place right now anyway, and it is sort of its own element in that its a more specific search result container so that seems fine as long as it hasnt had any affect on how it displays. I'll resolve this comment once the indentation/brackets have been tidied up.

app/assets/stylesheets/_search_result.scss
Rabid-Dan: @louise-r-blue flex can be a bit tricky because there are several related properties but here we're more specifically talking about a flex property rather than a display property with a value of flex. It can help you avoid a bunch of these floats and if you'd like to see an example have a look at how the "learn more" and "add to vocab" buttons are set up on the search results cards. That display logic looks to be mostly transferable on the face of it (both of them are set to flex: 50%)

app/assets/stylesheets/_search_result.scss
louise-r-blue: This was a bit of a wrangle and trying display of any kind wasn't perfect. Only worked best with the float position.

app/assets/stylesheets/application.scss
louise-r-blue: Yeah sorry about that one. Added back in now.

app/assets/stylesheets/_buttons.scss
louise-r-blue: @Dan @danieljauch without changing existing styles which include float properties the most accurate and flexible method seems to be retaining the float position property.

app/assets/stylesheets/_buttons.scss
Rabid-Dan: nah thats a pretty good call on our part. Happy to leave them as px for now for clarity.

app/assets/stylesheets/_buttons.scss
Rabid-Dan: @louise-r-blue did you take a look at the way the buttons on the search result cards are set up? They dont appear to use any floats and are basically the exact same thing at a smaller size.

app/assets/stylesheets/_search_result.scss
louise-r-blue: @Rabid-Dan Do you mean the one liner when you refer to the indentation?

app/assets/stylesheets/_buttons.scss
louise-r-blue: @Dan yes I my initial intention was to use it as a template but there is a lot more nesting happening in the search results page. I think the only way to emulate is to add more to the haml file?

louise-r-blue: And I didn't copy and paste anything so what is implemented on the sign details page didn't pick up styles from a different file.

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -4,6 +4,10 @@
       = hidden_field_tag "item[#{vocab_sheet_item.id}][id]", vocab_sheet_item.id, class: 'item_id'
       = link_to sign_url(vocab_sheet_item.sign_id), { class: "drawing vocab_sheet_drawing drawing vocab_sheet_drawing__size--#{size}" } do
         = image_tag "/images/signs/400-400/#{vocab_sheet_item.drawing}"
+        = link_to "/images/signs/400-400/#{vocab_sheet_item.drawing}" do
+          %span.icon-container
+            = image_tag "/assets/ui/download-button.svg/", class: "vocab-download-button"
```
Br3nda: is this uri supposed to end with `/` ?

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -4,6 +4,10 @@
       = hidden_field_tag "item[#{vocab_sheet_item.id}][id]", vocab_sheet_item.id, class: 'item_id'
       = link_to sign_url(vocab_sheet_item.sign_id), { class: "drawing vocab_sheet_drawing drawing vocab_sheet_drawing__size--#{size}" } do
         = image_tag "/images/signs/400-400/#{vocab_sheet_item.drawing}"
+        = link_to "/images/signs/400-400/#{vocab_sheet_item.drawing}" do
```
Br3nda, line 7: should we use `image_url` to hook into the asset pipeline here?

app/views/signs/show.html.haml
```diff
@@ -16,10 +16,15 @@
       .glosses-container.glosses
         %h2.main_gloss= @sign.gloss_main
         %h2.main_gloss.maori-gloss= @sign.gloss_maori
-      = link_to "#", { class: "card__button add-to-vocab-btn card__button--peach", "data-sign-id": @sign.id } do
-        %span.icon-container
-          %i.fi-plus
-        %span.text-container Add to Vocab Sheet
+      .button-container
+        = link_to "/images/signs/400-400/#{@sign.drawing}", { "data-sign-id": @sign.id } do
```
Br3nda: the path `/images/signs/400-400` is used a few times. Can it be in a variable or helper, some other way to DRY so it's not repeated.

Br3nda: something smart and short, such as `link_to @sign.image` or `sign_image_path(@sign)`

app/views/vocab_sheets/_vocab_sheet_item.html.haml
louise-r-blue: well spotted!

app/views/vocab_sheets/_vocab_sheet_item.html.haml
louise-r-blue: Thanks Brenda. I will check this out : }

Br3nda:
Two fixed issues in codeclimate.com. Nice!

app/views/vocab_sheets/_vocab_sheet_item.html.haml
louise-r-blue: Hi @Br3nda I ran this by Dan and he is not 100% sure but did provide this context for us. : }
Re Image routing:
The urls are routed to a specific controller action rather than served through the traditional asset pipeline.

Does that help?

Br3nda (approved):
Love your work!

app/assets/stylesheets/_buttons.scss
```diff
@@ -69,6 +68,46 @@ a .button,
...
+  left: 0;
+  text-align: left;
+  width: 100%;
+  .add-to-vocab-btn {
+    display: flex;
+  }
+}
+
+.download-link {
+  width: 50%;
```
Rabid-Dan, line 81: could this be changed to flex:50% please so it stays consistent with the other flex properties

app/assets/stylesheets/_buttons.scss
```diff
@@ -69,6 +68,46 @@ a .button,
...
+.download-button,
+.add-button-icon {
+  bottom: -4px;
+  height: 2.0625rem;
+  margin: 0 2px 0 0;
+  position: relative;
+  width: 2.1875rem;
+}
+
+.add-button-text {
```
Rabid-Dan: You can delete this add-button-text class if you also remove the float-right property thats been assigned on line 58 of _search_result.scss below

app/assets/stylesheets/_buttons.scss
louise-r-blue: @Rabid-Dan Done!

app/assets/stylesheets/_buttons.scss
louise-r-blue: @Dan I have removed the right float. All ok then when removing the 2.5rem left padding means the text floats over the button.

app/assets/stylesheets/_buttons.scss
Rabid-Dan: @louise-r-blue I'm not seeing the change? It still displays as width 50%

Rabid-Dan (approved):
Great stuff! Thanks for making all those changes!

PR #376, opened 2018-10-01
Fix db:seeds

Br3nda:
Rails 5 no longer allows `belongs_to` to parent records that don't exist.
At some point, some strange test data made its way into page_parts.csv, that have no matching record in pages.csv

 so the `rake db:seed` no longer works. This removes a bunch of test data from in there and gets it working again.


Br3nda:
Looks like the broken data was added by someone named Brenda in this commit 2fae7123708d8768667dad8b1a7ce27b4faed2d7

Rabid-Dan (approved):
👏 👏 👏

PR #390, opened 2018-10-03
Create Sitemap and store freelex xml data dump required for sitemap output

Rabid-Dan:
As part of the wider story of SEO optimisation I've generated a sitemap with a new controller and some assisting models.

Because signs require a data dump from freelex, I've opted to create a small, single attribute (for storing xml data) model for a sitemap as well as a supporting sitemap builder class that houses all of the logic required to return generate and update the sitemap model. I've also added a rake task to be run nightly that pulls in a fresh copy of the database dump and refreshes the sitemap data.
With any luck this extra infrastructure will stop the platform from being swamped with long running processes (which would otherwise occur every time the sitemap page was hit). Fetching the database dump for use in the sitemap takes a while so I've had to increase the faraday timeout limit to allow time for the task to run.

One last notable change is the shifting of the app domain name and protocol variables from the secrets file into the application config, and the establishing of a base url variable.


app/models/sign.rb
```diff
@@ -103,6 +103,7 @@ def self.query_string_for_search(query)
   def self.http_conn
     Faraday.new(url: SIGN_URL) do |faraday|
       faraday.use FaradayMiddleware::FollowRedirects
+      faraday.options[:timeout] = 60
```
elspeth-rabid, line 106: Out of curiosity, whet prompted the addition of this line ^^^?

lamorrison:
Hmm, if the sitemap takes a long time to build maybe it shouldn't be triggerable from a public controller action? It might be more appropriate for the sitemap controller to 404 if the sitemap hasn't been generated yet.

app/models/sitemap.rb
```diff
@@ -0,0 +1,2 @@
+class Sitemap < ApplicationRecord
+end
```
ghostPepperWonton, line 2: Is there a reason for this file?

Rabid-Dan: Its just the model that the sitemap xml data gets stored in. The SitemapBuilder class is more like a service class that houses all of the methods required to put some data into this.

Rabid-Dan:
@lamorrison thats a super good point that I hadn't really thought about. I think I agree but I'm not sure what sort of implications that has for SEO if the sitemap gets deleted for some reason and not re generated straight away. I'm pretty on the fence about the tradeoffs but maybe I could have it throw a raygun error if it 404s so at least we get an action to fix it rather than let the site quietly fall down search rankings?

louise-r-blue:
amazing !!! Great work

louise-r-blue (approved):

lamorrison:
> @lamorrison thats a super good point that I hadn't really thought about. I think I agree but I'm not sure what sort of implications that has for SEO if the sitemap gets deleted for some reason and not re generated straight away. I'm pretty on the fence about the tradeoffs but maybe I could have it throw a raygun error if it 404s so at least we get an action to fix it rather than let the site quietly fall down search rankings?

I think it's only an issue if regenerating the sitemap slams the site hard enough that it doesn't work properly for regular users, otherwise I agree that regenerating it on demand is better. I'll leave it to your conscience. Have fun! *saunters away, whistling nonchalantly*

lamorrison:
Oh, one more thing. If the sitemap takes a long time to generate you probably need to set a flag while it's generating so that another request won't start another process to generate another sitemap at the same time.

ghostPepperWonton (approved):

app/models/sitemap_builder.rb
```diff
@@ -0,0 +1,35 @@
+class SitemapBuilder
+  def first_or_generate
+    Sitemap.first || Sitemap.create(xml: generate_xml(page_slugs + sign_slugs))
+  end
+
+  def update_sitemap
+    Sitemap.first.update(xml: generate_xml(page_slugs + sign_slugs))
```
elspeth-rabid, line 7: 🐘 since you will always be calling for both page and sign slugs, and never individually, it seems worthy of being a method to me... even though the method is currently only combining 2 arrays.

app/models/sitemap_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+    end
+
+    builder.to_xml
+  end
+
+  private
+
+  def fetch_data_dump
+    Sign.all(xmldump: 1)
```
elspeth-rabid, line 25: What is this doing? What is this `xmldump` attribute? I sense this could have a better name... 🐘

app/models/sitemap_builder.rb
```diff
@@ -0,0 +1,35 @@
+class SitemapBuilder
```
elspeth-rabid, line 1: 🐘 is this a service or a model?

spec/models/sitemap_builder_spec.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
```
elspeth-rabid, line 1: ❤️ these tests - really clear 😄 

Is it testing whatever is up with the `xmldump` business?

elspeth-rabid (commented):
This is really good - so I turned the fussiness dial up 😺

app/models/sign.rb
Rabid-Dan: Fetching the datadump from freelex takes quite a while and leads to faraday timeouts if you try to grab the whole thing at once. This is toootally a compromise but it appears to be the cleanest way to get all of the signs from that datadump we need at the same time for the sitemap.

app/models/sitemap_builder.rb
Rabid-Dan: It was going to be a service, but then I had a good chat to josh about the semantics of service vs builder and because its a sitemap **builder** it was marginally more appropriate to make it a model.

app/models/sitemap_builder.rb
Rabid-Dan: I've just changed this in the first or generate method above, do you still reckon itd be worth combining these two in a wrapper method or something though?

app/models/sitemap_builder.rb
Rabid-Dan: The xmldump attribute is something that we need to pass in a call to freelex to get a collection of information for all of the signs stored in it. Theres a big ol set of methods in the sign model that handle it that I don't really want to touch too much, but maybe this could be called something like pull_sign_data_dump_from_freelex instead?

spec/models/sitemap_builder_spec.rb
Rabid-Dan: It tests the output of the sitemap builder methods based on a stubbed version of the data we expect to get back, but not the call itself. The call takes a wee while so testing the actual return value of the xmldump so it would make the tests take waaayyy longer and it would need to be a test on the Sign model anyway.

Rabid-Dan:
@elspeth-rabid thanks! Fussiness is much appreciated and super helpful!

Rabid-Dan:
@lamorrison I found a compromise to your comment. I've changed the output of the first or generate method to just create a basic sitemap with a few internal page slugs instead of the whole thing. That makes it suuuper quick and means that there'll always be something to display for web crawlers. When the rake task runs overnight it will update it with the full collection of sign data as well so it feels like a win win.

app/models/sitemap_builder.rb
elspeth-rabid: I would call the method `fetch_all_signs` then maybe?

app/models/sitemap_builder.rb
elspeth-rabid: It seems nice to have a method to call and know you will get ALL pages, but very much an 🐘, not important.

app/models/sign.rb
elspeth-rabid: I thought so, just checking.... it could be that you made an optional argument to this that allows you to extend the timeout in specific scenarios but totally 🐘

Rabid-Dan: yeah we did discuss something like that but its tied into enough existing code that doing that in a nice way would be lootts of extra work.

lib/tasks/update_sitemap.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+namespace :sitemap do
```
elspeth-rabid, line 3: 🌟 

Are you using the Heroku scheduler thingy to do this weekly or something?

elspeth-rabid (approved):
👍

lib/tasks/update_sitemap.rake
Rabid-Dan: yep thats the plan! originally I was going to do it nightly but thats probably overkill for something like this.

app/models/sign.rb
elspeth-rabid: I think an optional argument could be added with almost no knock on effect to the codebase:
https://holgerjust.de/2016/detecting-default-arguments-in-ruby/

```
def self.http_conn(timeout = 10)
  Faraday.new(url: SIGN_URL) do |faraday|
    faraday.use FaradayMiddleware::FollowRedirects
    faraday.options[:timeout] = timeout
  end
end

elspeth-rabid: But 🐘 !

PR #399, opened 2018-10-03
Ammend download icon image path and hide download icon in print view

Rabid-Dan:
Fixes a few small issues with image download icons display.

Changes the paths of where the icon is located to be correct for compiled version on staging.
Hides icon and its associated link wrapper in the print and pdf views.

I've also snuck in a quick fix for something I noticed in the vocab sheet sidebar where extra wide images were leaking out the right hand side slightly.


ghostPepperWonton (approved):

app/assets/stylesheets/print.scss
```diff
@@ -59,6 +59,12 @@ footer .copyright {
       }
     }
   }
+
+  .vocab_sheet_drawing{
```
elspeth-rabid: Linting ^^^

elspeth-rabid (approved):
:-)

louise-r-blue (approved):

app/assets/stylesheets/print.scss
Rabid-Dan: 🌟

vinithanatarajan (approved):

PR #400, opened 2018-10-04
remove download buttons from signs show and vocab sheet pages

Rabid-Dan:
The high resolution image downloading story wont be ready to go for another wee while yet because new images need to be uploaded. So as not to undo too much of the work thats been put into this functionality so far, this PR removes just the buttons from the views so that we can merge staging into prod for a final deploy.

The next feature branch for this functionality can revert this commit and continue with the new addtions


louise-r-blue (approved):
Thanks

ghostPepperWonton (approved):

vinithanatarajan (approved):

PR #405, opened 2018-10-07
Feature/create contributing doc

louise-r-blue:
This is the first PR for the contributing markdown file which includes the code of conduct. Would be good to get this up as the project is now open source. Please check esp the areas (typos and language) before and after the code of conduct.


ghostPepperWonton (approved):
Super minor thing you can choose to address: Some of the paragraphs are capped columns (looks like 80 columns), but some are as long as the line you have. There are some plugins to autoformat that and make it consistent. Looks good though!

louise-r-blue:
@danieljauch what do you think is the best/optimal paragraph width?
@danieljauch I added to the editor config but not sure there is any diff.

.editorconfig
```diff
@@ -11,3 +11,4 @@ trim_trailing_whitespace = true
 insert_final_newline = true
 indent_style = space
 indent_size = 2
+preferred_line_length = 80
```
Br3nda: i've been using 120 --

80 is a bit of an old standard, and it's rare to find a screen that small now.
:elephant:

Br3nda: We should match our rubocop settings too -- so actually 120 is a better idea i think

Br3nda (changes requested):
editor config as 120 would be better -- or at least match it to the rubocop linter config

Br3nda:
the contributor docs looks fabulous. maybe the editor config change should be on a different PR ?

.editorconfig
louise-r-blue: Ok I have removed line 14 from the config.

louise-r-blue:
@Br3nda all done I think

Br3nda (approved):

PR #406, opened 2018-10-08
Update/add header for seo

louise-r-blue:
This PR adds the nzsl canonical url ( as an environmental variable ) to the head section of the main application file. 

This will tell web crawlers which page it is.

What is a canonical URL: the search engine friendly URL that you want the search engines to treat as authoritative. In other words, a canonical URL is the URL that you want visitors to see. http://www.jodynimetz.com/what-is-a-canonical-url

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -8,7 +8,7 @@
       %meta{content: "noindex", :name => "robots"}/
     %meta{content: "#{t("layout.meta_description")}", name: "description"}/
     %meta{content: "IE-edge,chrome=1", "http-equiv" => "X-UA-Compatible"}/
-    %link{href: "https://www.nzsl.nz", :rel => "canonical"}/
+    %link{href: Rails.application.config.base_url, rel: "canonical"}/
     - if protect_against_forgery?
       %meta#authenticity-token{content: form_authenticity_token, name: "authenticity-token"}/
     = render partial: "shared/google_tag_manager_head"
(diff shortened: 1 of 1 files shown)


joshmcarthur:
@louise-r-blue is there an environment variable or similar configuration that you can get the canonical host from? I recall @Rabid-Dan's sitemap work had a config setting you might be able to use?

app/views/layouts/application.html.haml
```diff
@@ -8,6 +8,7 @@
       %meta{content: "noindex", :name => "robots"}/
     %meta{content: "#{t("layout.meta_description")}", name: "description"}/
     %meta{content: "IE-edge,chrome=1", "http-equiv" => "X-UA-Compatible"}/
+    %link{href: "https://www.nzsl.nz", :rel => "canonical"}/
```
Br3nda: can that be `rel: 'canonical'` - it's the way to write hashes in ruby 2.

Br3nda (changes requested):
new ruby syntax would be groovy

app/views/layouts/application.html.haml
Rabid-Dan: Like @joshmcarthur mentioned, we've got a base_url config variable setup for the href part of this now. It can be changed to Rails.application.config.base_url instead of "https://www.nzsl.nz"

Rabid-Dan (changes requested):

app/views/layouts/application.html.haml
louise-r-blue: Ok thanks

Rabid-Dan (approved):

KatieWright26 (approved):

Br3nda (approved):

PR #415, opened 2018-10-15
Auto corrected by following Ruby MagicComment

Br3nda:
Auto corrected by following Ruby MagicComment

Click [here](https://awesomecode.io/repos/ODNZSL/nzsl-online/ruby_config_groups/748) to configure it on awesomecode.io

### DIFF
--- a/app/models/rendered_pdf.rb
+++ b/app/models/rendered_pdf.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class RenderedPdf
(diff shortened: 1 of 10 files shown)


app/models/rendered_pdf.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
Br3nda: needs an ignore added to rubocop.tml for this file

PR #419, opened 2018-10-17
NZSL / download high resolution drawings

louise-r-blue:
This PR is for the requirement for an implementation that a user can download high resolution drawings from the sign details page OR from their own vocab sheets. 

- currently we expect all drawing images have the `default` text appended. for example the one entry for watermelon is the sign drawing for watermelon `4237/watermelon-4237-default.png`
- does it matter if there are 1, 2 or more instances of `default` in the drawing sign string? No, as we are interested in the last occurrence with the .png file extension.
- if there are drawing signs like `1234/melon-default.png`  or `1235/default-melon-default.png` or `1236/melon-default-default.png` or `1234/melon.png` only the sign drawings that include `default.png` will have the `default.png` exchanged for `high_resolution.png`.
- Any drawings without `default.png` can still be downloaded at the original resolution.
- If for any reason a high resolution drawing can not be downloaded the user will get the default drawing
- An option was suggested that the high resolution images should be served from the Freelex database so not to slow the web application. As the product owner wishes the downloading of drawings to be available off-line (like for schools) then the `default` drawings will be available to the user.
- when it comes to the story being implemented that users may be able to upload their own drawings, I believe this would come with a different set of logic - like validating file path, validating the drawing, checking the file extension, adding `default` etc:
- tests have been updated to reflect the renamed drawings with `default` appended to the file path.

Further to implementing the download high resolution drawing this PR also addresses the closely related issue when the high resolution image path with file extension  was being rendered underneath the drawing in the PDF and print views. Although we have styled the download button to `display: none;` in the print view the image url was being rendered when the drawing was printed off or downloaded as a PDF. An existing class was added to the download button in vocab sheet view to add a style so the path would not be rendered.

Finally a small Raygun error reported that the favicon icon could not be found. In discussion with @Eoin and @Rabid-Dan it was decided the path to the favicon was incorrect. At the same time I had requested a new favicon icon from the product owner who mentioned that the icon needed updating anyway. A note to this is users will not immediately see the new favicon because of browser caching.

When this PR is reviewed and after the product owner has tested in staging there will be some monitoring of the production logs (@jakerabid has added new logging scripts)  to help resolve further Raygun  issues.


ghostPepperWonton:
A potential issue I just thought of for this: what if the filename includes the word "default". The filename for example would be "1234_default-default" which would convert to "1234_high_resolution-high_resolution". Might it be better to split the string by the hyphen and change the string that _exactly_ matches "default" from the resulting array?

louise-r-blue:
@danieljauch I just did a check for 'default' and none comes up nor does high resolution but a very good point none-the-less.

Rabid-Dan:
I think that removing the download icon from the print/pdf views should be a requirement for this story, rather than an "if time allows" thing. If we know about it and still allowing it to be merged then we're willingly merging in a bug/unwanted change of behaviour.

spec/helpers/sign_download_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SignsHelper, type: :helper do
+  drawing_1 = "1234/a_test-default"
+
+  describe '#convert_to_high_resolution' do
```
Rabid-Dan, line 12: ⭐️

spec/helpers/sign_download_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    context 'when provided with a "default" sign drawing' do
+      it 'returns the sign drawing with the text "high_resolution" as part of the file name' do
+        expect(helper.convert_to_high_resolution(drawing_1)).to include("high_resolution")
+      end
+    end
+  end
```
Rabid-Dan, line 42: maybe an additional test for this helper to assert that if "default" isn't present in the sign name then it nothing is modified and the returned value is the same is what gets passed in?

Rabid-Dan:
@louise-r-blue I think I agree with @danieljauch that the logic might need to change. We can ask Micky to reserve the word but theres no guarantee that she'll be the one uploading these for the rest of the sites life span. Adding a marginally more complicated solution to split on hyphens and then replace the last instance of "default" with "high-resolution" before rejoining them feels like a more forward thinking way to avoid any human errors.

ghostPepperWonton:
> I will let Micky know that she will have to reserve these words.

@louise-r-blue Ideally, there would be an implementation that allows for any word in the database since it is a dictionary after all.

spec/helpers/sign_download_helper_spec.rb
louise-r-blue: @danieljauch @Rabid-Dan yes I agree with you both. I won't be making changes today but it will get done. Cheers

louise-r-blue:
> > I will let Micky know that she will have to reserve these words.
> 
> @louise-r-blue Ideally, there would be an implementation that allows for any word in the database since it is a dictionary after all.

@danieljauch Yes you are right and more sensible than me. I have since changed the method that converts the `default` text to `high_resolution.png` by finding the last occurrence of `default.png`. So if we have a sign drawing example that is `4422/icon-default-4422-default.png` a user can download `4422/icon-default-4422-high_resolution.png` I have allowed the user to download a default drawing if for some reason the high resolution drawing does not exist. So if `4422/icon-default-4422-high_resolution.png` doesn't exist the user gets `4422/icon-default-4422-default.png`

spec/helpers/sign_download_helper_spec.rb
louise-r-blue: Comment moved to top^^

app/views/signs/show.html.haml
```diff
@@ -17,6 +17,10 @@
         %h2.main_gloss= @sign.gloss_main
         %h2.main_gloss.maori-gloss= @sign.gloss_maori
       .buttons-container
+        = link_to "/images/signs/400-400/#{convert_to_high_resolution(@sign.drawing)}", { class: "download-link", "data-sign-id": @sign.id } do
```
Br3nda, line 20: questions -- is this a cache of images from Freelex server? What if the file isn't there yet? What if the file doesn't exist?

app/views/vocab_sheets/_vocab_sheet_item.html.haml
```diff
@@ -4,6 +4,10 @@
       = hidden_field_tag "item[#{vocab_sheet_item.id}][id]", vocab_sheet_item.id, class: 'item_id'
       = link_to sign_url(vocab_sheet_item.sign_id), { class: "drawing vocab_sheet_drawing drawing vocab_sheet_drawing__size--#{size}" } do
         = image_tag "/images/signs/400-400/#{vocab_sheet_item.drawing}"
+        = link_to "/images/signs/400-400/#{convert_to_high_resolution(vocab_sheet_item.drawing)}" do
```
Br3nda: this repeats html from show.html.haml. How about making something re-usable that renders the images?

app/views/signs/show.html.haml
Br3nda: Heroku spins up the app on many servers - you might run the fetching on one server, and then the download high res runs on a different on that doesn't have this cached -- so you'll get 404s that you won't get on dev, and it'll be confusing.

There's a sign_image_controller that you can image_tag to instead. That looks for if a file already exists locally, and fetches from freelex if it doesn't. Could be really helpful to prevent future devs being mystified.

Br3nda (commented):
not sure this will serve files okay on heroku or when not cached yet.
I suggest clearing that image cache dir in dev to see what i mean.

app/views/signs/show.html.haml
louise-r-blue: I am not totally 100% sure if this will be a cache of images from Freely server and yes you are right there needs to be some error handling to respond to the user if they are unable to download a image whether high_resolution or not. Currently the high_resolution images are not there as I think it was going to take a few weeks for Micky to organise.

app/views/vocab_sheets/_vocab_sheet_item.html.haml
louise-r-blue: Just confirming that this (line 7) doesn't render drawings. It is a download link so a user can download the drawing to their own computer.

app/views/signs/show.html.haml
Br3nda: something like `sign_image/show/000.png?hires=1` could work

Br3nda: it might be a better idea to link to the freelex high res images directly - so we don't fill up our disk space on the heroku servers.

DSRU:
All images have now been updated in the assets. The code is ready to be tested and then merged.

app/helpers/signs_helper.rb
```diff
@@ -20,6 +20,10 @@ def render_grammar_notes(sign)
     end.compact.join(', ').html_safe
   end
 
+  def convert_to_high_resolution(sign_drawing)
+    sign_drawing.gsub(/default.png$/, 'high_resolution.png')
```
Rabid-Dan: ooh nice change!

spec/helpers/sign_download_helper_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  describe '#convert_to_high_resolution' do
+    context 'when provided with a "default.png" sign drawing' do
+      it 'returns the sign drawing with the text "high_resolution" as part of the file name' do
+        expect(helper.convert_to_high_resolution(drawing_1)).to include("high_resolution.png")
+      end
+    end
+  end
+
+  describe '#stay_with_original_resolution' do
```
Rabid-Dan: just a heads up that the convention of `#name_of_a_thing` for a describe block is usually used to describe separate methods rather than scenarios, so you can probably get rid of this describe and the `'#convert_to_high_resolution and retain the string - default'` one below and just shift both of the context blocks into your `describe '#convert_to_high_resolution'` block above

app/assets/javascripts/vocab_sheet.js
```diff
@@ -41,13 +41,17 @@ $(document).ready(function() {
       $('input.vocab_sheet_name').blur(function() { submit_vocab_sheet_name($(this)); });
 
       if (document.printView) {
+        $('img.vocab-download-button').remove();
```
Rabid-Dan: I think the button this should be hidden with css in the print view rather than removed with javascript. The project already has a print specific style sheet so it should just be a case of adding something like a display:none for this class in that file. That way that stylesheet stays a single source of truth for how a print view should be customised rather than having changes in multiple files.

app/helpers/signs_helper.rb
kasperite: FYI this logic won't work with edge cases eg `something-Default.png` or `something-default.PNG`. `sign_drawing.gsub(/default.png$/i, 'high_resolution.png')` is what you need. Consider this as 🐘 if case sensitivity is irrelevant.

Rabid-Dan (approved):

kasperite (approved):

PR #442, opened 2018-11-21
add extra logging to try and diagnose heroku's H12ing requests

JacobBriggsAckama:
This adds heroku's request id to all logging from rails, so we can see
what requests are causing the application to take over a minute to
respond and perhaps diagnose the root cause


Br3nda:
Pretty sure it's the attempted call to Message.count or Reading.count. -- in some view code that says something like `if sensor.message.size.positive?` and `if room.reading.size.positive?`
Might need an index -- might need to do something smarter for working out if the sensor has sent a message yet.

JacobBriggsAckama:
@Br3nda are you sure you are commenting on the right project :P

Br3nda:
Woah! Totally spaced that

Br3nda:
There used to be code that logs how long freelex takes to respond. That could b3 logged again.

JacobBriggsAckama:
I might go and find and/or recreate that logging code then and update the PR. The freelex maintainer insists its not freelex causing this though and I am inclined to believe them tbqh and I'd love to rule it out, or in.

JacobBriggsAckama:
I don't know if the logging is good or not, but, it would be awesome if it was and this could be deployed to prod before the time-outs happen again and maybe Monday I can look better at this - I don't work fridays....

Br3nda (approved):

app/services/image_processor.rb
```diff
@@ -16,16 +16,24 @@ def initialize(filename:, height:, width:)
...
 
   private
 
-  def write_file_locally
-    image = MiniMagick::Image.open(@remote_filename)
+  def write_file_locally # rubocop:disable Metrics/AbcSize
+    image = nil
+    image_retrieval =
+      Benchmark.measure("retriving image '#{@remote_filename}'") { image = MiniMagick::Image.open(@remote_filename) }
+
```
Br3nda, line 30: Spelling mistake here

Br3nda:
Also worth having a look at  #73

app/services/image_processor.rb
JacobBriggsAckama: opps oh no

PR #465, opened 2018-12-17
Change download btn path

louise-r-blue:


louise-r-blue:
Sorry @Br3nda @Rabid-Dan GitHub is being fussy about an svg file which is fine locally but not showing in heroku staging so I have altered the path to test in staging.

app/views/signs/show.html.haml
```diff
@@ -19,7 +19,7 @@
       .buttons-container
         = link_to "/images/signs/400-400/#{convert_to_high_resolution(@sign.drawing)}", { class: "download-link", "data-sign-id": @sign.id } do
           %span
-            = image_tag "/assets/ui/download-button.svg", class: "card__button download-button"
+            = image_tag "/assets/images/ui/download-button.svg", class: "card__button download-button"
```
Rabid-Dan: I dont think this will work @louise-r-blue for an image_tag I think your url needs to just be "ui/download-button.svg" instead because the assets/images bit is handled for you. It would also explain why you see it locally right now but not remotely because "/assets/ui/download-button.svg" also wont compile the path correctly

Rabid-Dan (changes requested):

app/views/signs/show.html.haml
louise-r-blue: thanks @Rabid-Dan just like the 'apple-touch-icon' !!

vinithanatarajan (approved):

JacobBriggsAckama (approved):

Rabid-Dan (approved):

PR #478, opened 2018-12-21
Fix bug where invalid feedback was causing an exception to be raised.

eoinkelly:
An exception was being raised whenever a user submitted an invalid feedback form because the `@feedback` ivar was not being set properly in this case.

This change

* fixes that bug
* sets up system tests in the app (Rails recommends system tests since 5.1. They can happily live beside existing feature tests but most (future) Rails devs will be more familiar with them so I thought it was better to bite the bullet)
* add some system tests for the feedback form

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
 before_script:
 - gem install  --no-rdoc --no-ri 'brakeman' 'bundler-audit' 'pandoc-ruby'
 - nvm install node 9.10.0
-- npm install -g yarn@1.5.1
+- npm install -g yarn
 - yarn install
 - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
 - chmod +x ./cc-test-reporter
@@ -21,6 +21,7 @@ before_script:
 addons:
   apt_packages:
     - pandoc
+    - google-chrome-stable
   code_climate:
     repo_token: CC_TEST_REPORTER_ID
 script:
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -2,7 +2,7 @@
 
 require 'csv'
 
-def load(data_type)
+def load_data(data_type)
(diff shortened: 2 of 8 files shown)


eoinkelly:
Travis is failing because it's unhappy with the Chrome version. I'm looking into this.

db/seeds.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'csv'
 
-def load(data_type)
+def load_data(data_type)
```
eoinkelly, line 5: Had to rename this because it conflicted with Ruby's built-in `Object.load` when seeds were invoked from `Rails.applicaion.load_seed` 🤷‍♂️

eoinkelly:
Travis happy now

.travis.yml
```diff
@@ -13,14 +13,15 @@ env:
 before_script:
 - gem install  --no-rdoc --no-ri 'brakeman' 'bundler-audit' 'pandoc-ruby'
 - nvm install node 9.10.0
-- npm install -g yarn@1.5.1
+- npm install -g yarn
```
eoinkelly, line 16: Latest yarn is 1.12.3 - I'm not aware of a need to to lock it to a really old version so I changed this

spec/system/feedback_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'rails_helper'
+
+RSpec.describe 'User Feedback', type: :system do
+  before(:all) do
+    # These tests require the seed data to be present
+    Rails.application.load_seed
```
eoinkelly, line 6: Most other tests depend on a clean database (no seeds) so I'm loading seeds here. Loading seeds is also slow enough that we don't want to do it if we don't need to in a test run.

Br3nda: This app is older than npm. There is probably a better way to get yarn on travis by now.

Br3nda (approved):

PR #480, opened 2018-12-21
Upgrade Heroku stack

eoinkelly:
This app is on the `cedar-14` Heroku stack which is quite out of date. https://devcenter.heroku.com/articles/upgrading-to-the-latest-stack has details about how to upgrade.

I need a PR to exist so I can create a review app and check that the stack upgrade won't break anything.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,5 +1,6 @@
 {
   "name": "nzsl-online",
+  "stack": "heroku-18",
   "description": "Dictionary of New Zealand Sign Language",
   "scripts": {
     "postdeploy": "bundle exec rails db:migrate db:seed"
--- a/app.json
+++ b/app.json
@@ -59,9 +59,6 @@
     }
   },
   "formation": {
-    "worker": {
-      "quantity": 1
-    },
     "web": {
       "quantity": 1
     }


Br3nda:
@eoinkelly they will get billed for the time the review app runs, so if you leave it for a weekend of christmas, check that you remove the review app.

Br3nda:
i had a look and it seems to be working great

Br3nda (approved):

app.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "nzsl-online",
+  "stack": "heroku-18",
```
Br3nda, line 3: can't decide if this is good practise.
I've been pressing the "upgrade stack" button on my apps. Staging and then prod.

Br3nda (approved):

app.json
Br3nda: It is more explicit and dev-opsy. But it also can tie us to old ways in the future if we don't manage it.

eoinkelly:
Yep I had a click through and it seems to work fine. btw I'm _pretty_ sure there is no billing for the review app because it patterns off staging which is all on the free tier. I might be missing something but the Heroku dashboard seems to be telling me it's all $0

Br3nda:
Yeah, i think you're right

I have all my apps in a heroku team, so there is no free tier allowed.

But NZSL seems to be set up on personal accounts, so yeah, free.

app.json
eoinkelly: Hmmm yeah I understand that pov. I could just decline this PR and use the "Upgrade stack" button in the UI now that we are confident it will work. I'm happy to do either - you wanna make the call?

Br3nda:
Only last caution is -- if we deploy this now, who's gonna roll it back if it breaks over christmas?
If you're around to watch it, go ahead :-D

eoinkelly:
I'll be here today/tomorrow at least. My thought was that we would just leave it in staging for christmas and do a prod deploy next year when Micky is around but I don't have a heap of context on how the normal NZSL process works so I'm happy to do whatever you think works best.

Br3nda:
Sounds good. Might go earlier if a security patch is needed in the gems.

eoinkelly:
If a vuln is disclosed over the break I'll get notified for our other apps anyway and will create a PR directly against master here if need be.

PR #481, opened 2018-12-22
Use Unicorn on Heroku deployments

eoinkelly:
This PR completes the setup of Unicorn. There are some errors in Raygun related to some rails processes timing out waiting for a connection from the pool. I'm hoping this change will fix those too.

There is a review app for this change. I clicked around it and it seems to work fine. I will clean up the review app when this PR is closed.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 release: bundle exec rails db:migrate
-web: bundle exec rails server -p $PORT
+web: bundle exec unicorn -p $PORT -c ./config/unicorn.rb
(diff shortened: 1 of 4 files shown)


eoinkelly:
I think this is safe to go through to production but no worries if you would prefer to wait until the new year to merge it @Br3nda

Br3nda (approved):

Br3nda:
All good. I'm happy as long as you're the one who will keep an eye on it in case it needs to be undone.

eoinkelly:
Sweet, I'm happy to do that.

PR #482, opened 2018-12-23
Fix bug triggered by WEB_CONCURRENCY env var being a string not a number

eoinkelly:
When Heroku sets WEB_CONCURRENCY to a string representation of a number,
not an actual number e.g. "3" not 3. This change accounts for that properly
when configuring the DB pool size at application startup

This change also removes the `|| 5` fallback which will never be used because
the fetch of WEB_CONCURRENCY from the env has a default value fallback.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -23,7 +23,7 @@ default: &default
   # available to be able to run things like `rails console` without impacting
   # the worker processes.
   #
-  pool: <%= ENV["DB_POOL"] || (ENV.fetch('WEB_CONCURRENCY', 3) + 2) || 5 %>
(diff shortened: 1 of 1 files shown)


eoinkelly:
This was my bad, doh! 🤷‍♂️

Br3nda (commented):
What does the +2 do? What's that about?

eoinkelly:
The comment above it explains it but it isn't part of this diff - see https://github.com/ODNZSL/nzsl-online/blob/e01c9510c204be577a8035e1a92d78afc2660279/config/database.yml#L22-L26

Br3nda (approved):

PR #483, opened 2018-12-23
Improve error reporting of freelex comms issues

### LINKED ISSUE #383: NZSL - 28. Gracefully handling Freelex issues
This issue requires an implementation to 

- Handle timeout/disconnect and 
- Display a custom site error rather than a Heroku server error

eoinkelly:
This PR makes the following changes:

### Create Sign::FreelexCommunicationError exception class

A custom exception class will allow us to see all the exceptions related to this class in a single group in our exception reporting - currently those errors are spread across many exception classes making it it difficult to troubleshoot.

### Change Freelex timeout from 60 sec to 20 sec.

Heroku enforces a 30 second timeout (details: https://devcenter.heroku.com/articles/request-timeout) on our generating a complete response. We set a timeout for Freelex that allows us time to process the response and return it to our client within that constraint.

Before this change, timeout and connection delays were sometimes being reported as different exceptions e.g. `ActionView::Template::Error because the Heroku timeout would cause Rails to fail elsewhere in its processing.

closes #383

### DIFF
--- a/app/models/sign.rb
+++ b/app/models/sign.rb
@@ -79,7 +79,16 @@ class Sign
   def self.search(params)
     xml_request(params)
   rescue FreelexCommunicationError => e
+    msg = <<~EO_MSG
+      Recovered from a failure to retrieve data from Freelex!
+        * An empty result-set will be returned to the calling code.
+        * Error details:
+          #{e}
+    EO_MSG
+
+    Rails.logger.warn(msg)
     Raygun.track_exception(e)
+
     [0, []]
   end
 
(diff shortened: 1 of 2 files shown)


Br3nda (commented):
You got time to add dns look up fail? That happens to newbie devs when they havent configired, and it would be useful to have an error message to give them a hint.

app/models/sign.rb
```diff
@@ -103,7 +120,15 @@ def self.query_string_for_search(query)
...
       faraday.use FaradayMiddleware::FollowRedirects
-      faraday.options[:timeout] = 60
+
+      # Heroku enforces a 30 second timeout on our generating a response. We
+      # set a timeout for Freelex that allows us time to process the response
+      # and return it to our client within that constraint.
+      #
+      # Faraday timeout option specifies when how long we should wait (in
+      # seconds) for data to be available to be read from the socket
+      faraday.options[:timeout] = 20
```
Br3nda: Can this be in config?

eoinkelly: Yep good idea, will do.

eoinkelly:
I added a warning to the logs when we recover from a Freelex e.g. 

```
Recovered from a failure to retrieve data from Freelex!
  * An empty result-set will be returned to the calling code.
  * Error details:
    Failed to connect to Freelex at URL: 'http://nzsl-assets.vuwxx.ac.nz/dnzsl/freelex/publicsearch'
```

Devs see that log output in the terminal output of their `rails s` session so it should give them a hint that something is amiss. I also included the `SIGN_URL` in the exception message in cases where we have a `Faraday::ConnectionError` (that is the error which gets raised if there is a DNS issue. It should help folks notice that there is a problem.

Br3nda:
Awesome. I'll give it a test tonight.

Br3nda (approved):

Br3nda:
@eoinkelly I've added the magic GitHub phrase `closes #383` to the description at the top of this PR - that makes GitHub close it automagically when this merges. It also links the issue to this PR in waffle

Br3nda:
attn: i rebased this.

PR #484, opened 2018-12-24
Refactor autocomplete search and improve error handling & logging

eoinkelly:
* Pull autocomplete searching out into a service object
* Use Faraday instead of built-in `Net::HTTP` because `Faraday` is much more
  disciplined about the exceptions it will raise (they are all under `Faraday::Error`)
* Log a warning and return an empty result-set if anything goes wrong while
  doing the autocomplete search
* Log autocomplete requests in development (handy for debugging)

From my testing it looks like the autocomplete search at `http://nzsl-assets.vuw.ac.nz/dnzsl/ncbin/public_search_lookup` ignores the `limit=10` param we send e.g. this is what I get when the search term is `hea`

```
Received 36 autocomplete suggestions (26 more than expected):
  ["heal", "head", "heat", "hear", "heavy", "heart", "health", "heater", "heaven", "header", "headway", "healthy", "heating", "hearing", "heading", "head for", "headache", "headline", "head lice", "heartburn", "heartbeat", "heartache", "headmaster", "heart rate", "headphones", "hearing aid", "heartbroken", "heart attack", "head teacher", "headquarters", "head spinning", "heavy drinker", "hearing school", "hearing impaired", "hearing aid battery", "hearing aid amplifier"]
```

Are there any docs/code I can look at to see what the deal is there? I've worked around it by trimming the results in the service but it would be better to just ask for the correct number if possible.

### DIFF
--- a/app/services/autocomplete_search_service.rb
+++ b/app/services/autocomplete_search_service.rb
@@ -2,6 +2,8 @@ class AutocompleteSearchService
   MAX_NUM_SUGGESTIONS = 10
   AUTOCOMPLETE_SEARCH_TIMEOUT = 10 # seconds
 
+  class AutocompleteSearchServiceError < StandardError; end
+
   ##
   # @param [String] search_term
   # @param [Faraday::Connection] faraday_connection - has a sensible default value which can be overridden by test code
@@ -43,6 +45,7 @@ class AutocompleteSearchService
         * Error details:
           #{e}
     EO_MSG
+    Raygun.track_exception(AutocompleteSearchServiceError.new(msg))
     @logger.warn(msg)
 
     []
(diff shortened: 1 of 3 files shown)


app/services/autocomplete_search_service.rb
```diff
@@ -0,0 +1,63 @@
...
+  #
+  def find_suggestions
+    response = @faraday_connection.get do |request|
+      request.params[:limit] = MAX_NUM_SUGGESTIONS
+      request.params[:q] = CGI.escape(@search_term)
+    end
+
+    results = response.body.split("\n")
+
+    if results.length > MAX_NUM_SUGGESTIONS
```
eoinkelly, line 31: I'll remove this once we sort out what is happening with the `limit` param - I added it to help verify that the limit is currently being ignored.

Br3nda (approved):
works great.

eoinkelly:
I have moved the question about the `limit` into Issue #490 because I don't think it needs to block merging this work.

davemosk:
I've added the limit= param into the Freelex public_search_lookup script, and will let you know when it's been put into production.

Issue #384, opened 2018-10-02
NZSL - 23 CodeClimate cleanup


louise-r-blue:
Ideally, this would sweep the entire application and clean up all issues that CodeClimate has been bringing up.

More immediately, the work done on NZSL‌-14 brought out a lot of issues that went unaddressed for the sake of the sprint.

- CSS class names need to be converted to BEM style: `block-name__element-name--modifier-name`

- Places where css properties are overridden should be better arranged to allow the cascade to create priority so that we don't need `!important`‌s

- CSS classes are added to elements so that rules can be applied to views without needing vague or conflict-prone CSS rules

Br3nda:
Link to codeclimate issues: https://codeclimate.com/github/ODNZSL/nzsl-online

eoinkelly:
I collected some stats

```
Total issues:   1647
Ruby issues:   14
JS issues:       109
SCSS issues: 1524
```

eoinkelly:
Codeclimate issues update:

* Total issues: 512

made up of

* Ruby issues: 14
* JS issues: 109
* SCSS issues: 389

eoinkelly:
Codeclimate has 146 issues total now https://codeclimate.com/github/ODNZSL/nzsl-online/issues

Br3nda:
Nice work

G-Rath:
We're no longer using CodeClimate for this project

PR #485, opened 2018-12-25
Rubocop cleanup

eoinkelly:
This attempts to address issue #384 and have no rubocop issues by

1. Use the same version of rubocop locally and on codeclimate (0.60 is the most recent available on codeclimate)
1. Fix all rubocop existing issues except for one which will not be relevant once PR #484 lands.


eoinkelly:
Codeclimate seems to be complaining about

1. the `Kernel.open` issue which is fixed by another PR
2. that mildly terrifying pagination logic that I touched to make `rubocop` happy.  I don't want to get into refactoring that logic in this PR but I'm not sure what the convention for dealing with Codeclimate issues on this project is. Advice is welcome.

app/helpers/pagination_helper.rb
```diff
@@ -49,7 +49,7 @@ def pages(page, total_pages) # rubocop:disable Metrics/AbcSize, Metrics/Cyclomat
           nil
         elsif (page < 5 && i == 6) ||
               (page > total_pages - 4 && i == total_pages - 5) ||
-              (page > 4 && page < total_pages - 3 && (i == page + 2 || i == page - 2))
+              (page > 4 && page < total_pages - 3 && [page + 2, page - 2].include?(i))
```
Br3nda: I can't decide if this is easier to read. Well done chipping away at that file. It's a huge logic pile.

eoinkelly:
I'm biting the bullet and sorting `PaginationHelper` - I'll push in the next few days

app/helpers/application_helper.rb
```diff
@@ -46,12 +46,6 @@ def play_video_button(text, url = nil, options = {})
             { class: "button #{options[:class]}" }.reverse_merge(options)
   end
 
-  def query_for_query_string
```
eoinkelly, line 49: Moved logic into `SignPaginationService`

app/services/sign_pagination_service.rb
```diff
@@ -0,0 +1,168 @@
...
+  end
+
+  def wrap_in_li(content)
+    content_tag(:li, content)
+  end
+
+  def generate_elipsis
+    wrap_in_li(content_tag(:span, '...', class: 'a'))
+  end
+
```
eoinkelly, line 59: I left the original pagination logic intact (I just renamed things). It is is still very complex but tbh when I extracted it any further it just made it harder to understand.

eoinkelly:
I refactored `PaginationHelper` into `SignPaginationService` - it seemed complex enough to need more than a few methods (Modules can't have private methods so a service class seemed reasonable). I added tests which have expected output generated by the old `PaginationHelper` logic so (assuming my tests cover enough cases) there shouldn't be any change in behaviour after the refactoring.

app/controllers/signs_controller.rb
```diff
@@ -1,15 +1,18 @@
...
 
+class SignsController < ApplicationController
   before_action :find_vocab_sheet, :set_search_query, :footer_content
 
   def search
     search_query = process_search_query
     @page_number = permitted_params[:p].present? ? permitted_params[:p].to_i : 1
     @results_total, @signs = Sign.paginate(search_query, @page_number)
     @query = search_query
+    @pagination_html = SignPaginationService.new(current_page_number: @page_number,
```
Br3nda, line 13: Maybe this should be accessed by a helper for the HTML view, instead of a variable on the controller.

app/services/sign_pagination_service.rb
Br3nda: Way easier to understand

Br3nda (commented):
Haven't had time to run and check it quote yet.

Br3nda (approved):
looks much better -- and I tested and pagination still works.

Br3nda:
@eoinkelly don't forget to delete the branch after merging (unless you want to keep working on that branch)

PR #497, opened 2018-12-31
Removed unused translations and normalised the file

Br3nda:


config/locales/en.yml
```diff
@@ -1,153 +1,81 @@
-# encoding: UTF-8
-# Sample localization file for English. Add more files in this directory for other locales.
-# See https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting points.
-
+---
 en:
-  back: Back
+  admin:
```
eoinkelly, line 3: `admin.save` is still in use in https://github.com/ODNZSL/nzsl-online/blob/staging/app/views/admin/settings/edit.html.haml#L14

eoinkelly (changes requested):
LGTM other than `admin.save`

config/locales/en.yml
Br3nda: i think that's missing a called to t('admin.save'). I'm a bit surprised that works

Br3nda: AHAH! went searching on apidock.com for submit_button - but it's really a helper method in this app.

i now wonder if i've missed other things translated via  `submit_button` I'll need to check more before we merge

eoinkelly: `submit_button` calls `t()` on it - see https://github.com/ODNZSL/nzsl-online/blob/staging/app/helpers/application_helper.rb#L14-L18

UI shows:

screenshot 2019-01-01 09 52 54

PR #502, opened 2019-01-02
Auto corrected by following Lint Ruby RSpec/DescribedClass

Br3nda:
Auto corrected by following Lint Ruby RSpec/DescribedClass

Click [here](https://awesomecode.io/repos/ODNZSL/nzsl-online/ruby_configs/25441) to configure it on awesomecode.io


eoinkelly:
I don't have strong feelings on this but my $0.02 is that `described_class` is a bit too clever in test files - it means you only have one place to change if you renamed the class under test but tbh text editors make that pretty easy (especially given that all the uses of it are in one test file).

 I'm happy if you want to merge it @Br3nda - I just think it's not particularly an improvement. OTOH  keeping Awesomecode happy might be better for project health in the long run so happy to go with your call on this.

Br3nda:
It's used to DRY the code more. When we use `described_class` instead, we can start to pull repeating test code into shared code.

eoinkelly (approved):

Rabid-Dan (approved):

PR #530, opened 2019-01-19
scss_lint is no longer under active development so use sass-lint instead

eoinkelly:
scss_lint is not being developed any more - see notice at top of
https://github.com/brigade/scss-lint so use sass-lint instead because

1. it has a similar feature-set
2. it is under active development
3. there are some "auto-fix" tools for it which might be useful for us

https://github.com/sasstools/sass-lint

As well as configuring `sass-lint`, this PR includes the result of my running `sass-lint-auto-fix` (from NPM) across `app/assets/stylesheets`. This tool, along with some configuration of `.sass-yml` has reduced the number of `sass-lint` warnings from 2589 to 469 which, while still high, is not as terrifying.


eoinkelly:
This PR is related to issue #384

Br3nda (approved):

eoinkelly:
Added more fixes for whitespace issue.

PR #532, opened 2019-01-20
Allow all specs and code lints to be run locally

eoinkelly:
This change allows a developer to run the same lints and specs that Travis, Codeclimate and Deppbot will run. This allows us to find problems before pushing (so hopefully less noise for reviewers).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,6 +2,11 @@ module.exports = {
   extends: 'google',
   env: {
     node: true,
+    browser: true,
     es6: false
   },
+  rules: {
+    "no-var": "off",
+    "require-jsdoc": "off"
+  }
 };
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,11 +1,8 @@
 engines:
+  # https://docs.codeclimate.com/docs/eslint
   eslint:
     enabled: true
-    checks:
-      no-var:
-        enabled: false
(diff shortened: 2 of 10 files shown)


.eslintrc.js
```diff
@@ -2,6 +2,11 @@ module.exports = {
   extends: 'google',
   env: {
     node: true,
+    browser: true,
     es6: false
   },
+  rules: {
+    "no-var": "off",
+    "require-jsdoc": "off"
```
eoinkelly, line 10: These two rules were moved from .codeclimate.yml

Br3nda (approved):
Approved. Thank you for your mahi.

In other projects I've used the overcommit gem. It becomes a standard way to run linters, and remind devs to rerun them.

Rabid-Dan:
👏 👏 👏

.eslintrc.js
louise-r-blue: @eoinkelly This is such a sensible change. Will the README be updated to reflect the changes. Something like `scripts described here` to run locally should be executed locally before pushing up a branch.

eoinkelly: I did add to the README about this https://github.com/ODNZSL/nzsl-online#running-the-tests - is that what you mean?

louise-r-blue: Oh ok ! I'm a few steps behind.

PR #561, opened 2019-02-16
Install Rack::CanonicalHost middleware to redirect non-canonical hostnames

eoinkelly:
## Change summary

This application has multiple DNS names for historical reasons, namely:

* www.nzsl.nz (canonical)
* nzsl.vuw.ac.nz (legacy, some printed literature still refers to it)

For SEO and effective CDN caching purposes we want to have just one canonical DNS name but we also cannot break existing links which might use the other names.

Traditionally this problem would be solved via URL rewriting in Apache/Nginx/Varnish etc. but that is not possible because we are hosted on Heroku.

The `rack-canonical-host` middleware allows us to solve this problem in Rails. It redirects requests for non-canonical domain names to the canonical domain name.

`Rack::Sendfile` is the first middleware in the Rails middleware stack (run `bundle exec rails middleware` for details) so we are inserting the Rack::CanonicalHost middleware right at the start. This means we don't waste time processing a request which will be redirected by Rack::CanonicalHost anyway.

## Deployment considerations

I couldn't figure out a convenient way to write an automated test for this. I tested it locally and have taken the following steps to mitigate any problems with it going into staging and production

1. The canonical host redirecting can be disabled by deleting the environment variable and restarting the dyno so rollback will be quick if there is an issue
1. I have defined `CANONICAL_HOST` appropriately for both the production and staging envs
1. I have configured `Rack::CanonicalHost` to issue a `no-cache` Header so browsers should not cache the redirection. Once we are confident that the redirection works as expected I will make another PR to enable caching - see  https://github.com/tylerhunt/rack-canonical-host


eoinkelly:
BTW This build looks like it fails but it is because CodeClimate is complaining about test coverage. As I mentioned above, I can't find a convenient way to write an automated test for this and I can't find a way to tell CodeClimate to ignore this 🤷‍♂️

eoinkelly:
I have been tailing the logs on production and it seems like more traffic comes to `nzsl.vuw.ac.nz` than `www.nzsl.nz`. Any traffic which comes to `nzsl.vuw.ac.nz` completely bypasses the CDN. I think landing this change should fix that.

Br3nda (approved):
Tested on heroku and it doesn't break

PR #562, opened 2019-02-18
Install and configure NewRelic for application performance monitoring

eoinkelly:
NewRelic has a free-tier which keeps data for 24hrs which is long enough to help diagnose where an application spends its time.

This application is still slower than I want it to be. NewRelic will make it much faster to 

1. Find out where the app is spending its time
1. Find out which controller endpoints are slowest

I have created a NewRelic account ($0 for the free-tier) and put secrets in the usual place. I have added the license key to both Heroku environments.


eoinkelly:
Going to merge this because it seems straightforward. I will deal with any deployment issues immediately on staging

Br3nda:
Keep an eye on it - i found new relic very useful, but on the smaller heroku it can be a performance hit too.

eoinkelly:
@Br3nda Will do. I've already uncovered some useful stuff from staging i.e. the rails server isn't slow but is stuck behind serving files so requests to it seem slow. I'm totally fine with disabling it after a few weeks either way.

Issue #490, opened 2018-12-30
Limiting of autocomplete search results not working


eoinkelly:
From my testing it looks like the autocomplete search at `http://nzsl-assets.vuw.ac.nz/dnzsl/ncbin/public_search_lookup` ignores the `limit=10` param we send e.g. this is what I get when the search term is `hea`

```
Received 36 autocomplete suggestions (26 more than expected):
  ["heal", "head", "heat", "hear", "heavy", "heart", "health", "heater", "heaven", "header", "headway", "healthy", "heating", "hearing", "heading", "head for", "headache", "headline", "head lice", "heartburn", "heartbeat", "heartache", "headmaster", "heart rate", "headphones", "hearing aid", "heartbroken", "heart attack", "head teacher", "headquarters", "head spinning", "heavy drinker", "hearing school", "hearing impaired", "hearing aid battery", "hearing aid amplifier"]
```

This can be replicated by

```
$ curl  "http://nzsl-assets.vuw.ac.nz/dnzsl/ncbin/public_search_lookup?q=hea&limit=10"
heal
head
heat
hear
heavy
heart
health
heater
heaven
header
headway
healthy
heating
hearing
heading
head for
headache
headline
head lice
heartburn
heartbeat
heartache
headmaster
heart rate
headphones
hearing aid
heartbroken
heart attack
head teacher
headquarters
head spinning
heavy drinker
hearing school
hearing impaired
hearing aid battery
hearing aid amplifier
```

Are there any docs/code I can look at to see what the deal is there? Are we passing the limit param incorrectly? Does this endpoint actually support a limit?

This doesn't currently impact users of this app because we are trimming the results before sending them to the browser but it would be better to just ask for the correct number if possible.

eoinkelly:
Update: We have been in touch with the maintainer of Freelex. They are fixing this issue and will let us know when it hits production.

DSRU:
@eoinkelly on staging it does look like the autocomplete results are now limited to 10. Is it necessary to limit the results though? It is quite useful to show all possible results to encourage 'browsing' behaviour. If a limit is desirable, then could the top 10 autocomplete options be ordered in alphabetical order rather than by length?

eoinkelly:
I don't have strong opinion about the limiting. I implemented the limiting in the controller because we were passing `limit=10` to Freelex so I assumed limiting to 10 was desired behaviour.

Br3nda:
How about we set the limit to a much larger number?
[...]

PR #566, opened 2019-02-21
Improve autocomplete

eoinkelly:
Sort autocomplete results and increase limit

This change fixes issue #490. It:

* Increases the limit on autocomplete results from 10 to 100.
* Sort autocomplete results alphabetically
* Stop logging about getting too many results from Freelex because we don't care anymore.


eoinkelly:
@DSRU I have made the requested changes. I kept a limit because it will help protect the user from getting a big wad of text as autocomplete if there is ever a bug between us and Freelex. I have used it locally and it seems to work well (the sorting is a good improvement I think) but you can have a play when it hits staging and see if it does what you expect.

app/services/autocomplete_search_service.rb
```diff
@@ -45,20 +45,12 @@ def find_suggestions # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
...
-          #{results}
-      EO_MSG
-
-      @logger.info(msg)
-    end
-
-    results.take(MAX_NUM_SUGGESTIONS)
+    response
+      .body
+      .force_encoding(Encoding::UTF_8)
```
Br3nda, line 50: Do we still need this? Had it been fixed on freelex?

Br3nda (approved):

app/services/autocomplete_search_service.rb
eoinkelly: I haven't heard anything from Dave yet so I assume not. I'll ask again just in case.

DSRU:
Dave hasn't been able to get in to the university servers yet - he's been trying to sort it out with them.

________________________________

PR #572, opened 2019-02-23
Improve autocomplete sorting

eoinkelly:
Implement feedback from @DSRU about autocomplete search:

* Autocomplete results should be sorted case insensitively.
* Autocomplete should submit the search when the user chooses a value from the suggestions pop-up.


Br3nda (approved):

DSRU (commented):
Looks great!

PR #574, opened 2019-02-23
Use puma instead of unicorn as app server

eoinkelly:
Heroku now recommends `puma` as app server because it gets you a bit more
performance in memory constrained environments like Heroku - see
https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server

`puma` can be tuned with the following environment variables (see
`./config/puma.rb` for details):

    WEB_CONCURRENCY = number of worker processes to run
    RAILS_MAX_THREADS = number of threads each worker should run

The default values of each environment variable are set as per Heroku's
recommendation in
https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server

This change replaces `unicorn` with `puma` for production environments -
`puma` was already the default app server in development & test.


Br3nda (approved):

eoinkelly:
I have been testing this in a Heroku Review app.

When restarting the dyno, the logs show the following line:

```
2019-02-23T04:21:38.669921+00:00 app[web.1]: [4] ! WARNING: Detected 1 Thread(s) started in app boot:
2019-02-23T04:21:38.669963+00:00 app[web.1]: [4] ! #<Thread:0x0000557b1891de10@/app/vendor/bundle/ruby/2.5.0/gems/activerecord-5.2.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:299 sleep> - /app/vendor/bundle/ruby/2.5.0/gems/activerecord-5.2.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:301:in `sleep'
```
https://github.com/schneems/puma_worker_killer/issues/35 which includes folk who work at Heroku indicates that this warning can be safely ignored - see https://github.com/schneems/puma_worker_killer/issues/35#issuecomment-218275912

Otherwise this seems to work fine.

PR #575, opened 2019-02-23
Improve page speed

eoinkelly:
### Background

I have been looking at the data we are getting in NewRelic now about site speed and noticed some patterns

screenshot 2019-02-23 10 21 33

The graph above is an overview of server performance (timespan is last 5 hours ish). We can see that the server spent the majority of its time in  `Web external` (the green in the graph) i.e. the server spent most of its time making calls to Freelex - this is about what we would expect.

If we dig into which endpoints spent the most time then we see that our "show a page" endpoint dominates (see graph below). This is the endpoint which generates all the pages on the site that are not an individual sign e.g. Home, numbers, alphabet etc.

screenshot 2019-02-23 10 22 46

And when we dig into that one endpoint, we see that it spends the vast majority of its time sending requests to Freelex (the brown in the graph below)

screenshot 2019-02-23 10 23 09

Now I dug into which pages actually talk to Freelex and it turns out only two do:

1. The numbers page pulls all the required number signs from Freelex every time you refresh it.
2. The homepage pulls a random sign from Freelex everytime you refresh it.

### Why choose memcache for the Rails cache?

Rails caching (details [here](https://guides.rubyonrails.org/caching_with_rails.html#low-level-caching)) can use many different backends e.g. Memcached, Redis, Files, Memory. I chose Memcached. My reasoning was:

* Filestorage is reputedly slow on Heroku and our cache would be destroyed everytime we deployed or Heroku restarted the dyno (see https://www.speedshop.co/2015/07/15/the-complete-guide-to-rails-caching.html)
*  In-memory storage would increase the size of each worker process. This is not suitable for Heroku where we want to be as memory efficient as possible. It would also mean that each worker process would have its own cache
* Redis & Memcache - the free Memcached addon for Heroku gives us 30MB of cache space and the Redis addons only give 25MB so I picked Memcached for the extra 5MB. The advantages of Memcached/Redis is that the cache will persist across dyno restarts. The main disadvantage is that there is network latency between the dyno and the cache server. However I am very confident that this latency will be negligible compared to how much time we are saving by caching.

I have added free Memcached addons to each Heroku environment in preparation for this PR landing.

### Why this change?

My goal is to reduce server memory usage  so we can go from a 2x to 1x heroku dyno. To achieve this we need to reduce the number of worker processes we need. To reduce the number of worker processes we need to 

1. Send fewer requests to them (the CDN we setup recently is doing the job here)
2. Make each request process faster so the worker can handle more requests per minute

This PR is in service of point 2. above. 

### What does this change do?

This PR does the following:

1.  Caches the signs used to power the "Numbers" page for 7 days (so we only make one request to Freelex for numbers every 7 days)
1. Caches the _sign of the day_ on the homepage for 24 hours.
1. Re-implement the "Show me another sign" button on the homepage

These pages are very popular (particularly the homepage) so making them faster will free up a lot of "worker time".

### Future

There are other opportunities for caching in this app but, because caching does add a bit of complexity, I won't add caching unless I think the cost/benefit is worth the extra complexity.


app/helpers/numbers_helper.rb
```diff
@@ -2,16 +2,21 @@
 
 module NumbersHelper
   def numbers # rubocop:disable Metrics/AbcSize
-    return @numbers if @numbers.present?
+    Rails.cache.fetch('numbers', expires_in: 24.hours) do
```
eoinkelly: This PR caches the numbers for 24 hours. @DSRU can we cache them for longer? My assumption is that would almost never change so caching for multiple days seems feasible. We can always manually empty the cache if we needed to push a change through.

config/environments/development.rb
```diff
@@ -18,10 +18,13 @@
   # Run rails dev:cache to toggle caching.
   if Rails.root.join('tmp', 'caching-dev.txt').exist?
     config.action_controller.perform_caching = true
-    config.cache_store = :memory_store
+
     config.public_file_server.headers = {
       'Cache-Control' => "public, max-age=#{2.days.to_i}"
     }
+
+    config.cache_store = :file_store
```
eoinkelly: Enabling `:file_store` in development because it is much easier to work with - you can see exactly what is cached on disk.

app/helpers/numbers_helper.rb
DSRU: @eoinkelly yes, they can be cached for longer or even treatd as a static page and I can supply static images in the same way as is done for fingerspelling and classifier pages.

eoinkelly: I have created issue #576 to track making it a static page. For now I am going to increase the cached time to 7 days  .

eoinkelly:
I have some failing tests. Will look at them when I get a chance

Br3nda:
wow nice work!

app/controllers/pages_controller.rb
```diff
@@ -9,6 +9,16 @@ def show
 
     @title = @page.title
     @feedback = Feedback.new if @page.template == 'feedback'
+    @sign = Sign.sign_of_the_day
```
Br3nda, line 12: :panda_face: :+1:

config/environments/development.rb
```diff
@@ -18,10 +18,13 @@
   # Run rails dev:cache to toggle caching.
   if Rails.root.join('tmp', 'caching-dev.txt').exist?
     config.action_controller.perform_caching = true
-    config.cache_store = :memory_store
+
     config.public_file_server.headers = {
       'Cache-Control' => "public, max-age=#{2.days.to_i}"
     }
+
+    config.cache_store = :file_store, "#{Rails.root}/tmp/rails-cache"
```
Br3nda, line 26: Is this going to cache successfully on Heroku? New dynos can spin up without any cache files. The database might be a better place.

Br3nda: I've used the db cache quite successfully on https://www.growstuff.org

config/locales/en.yml
```diff
@@ -22,7 +22,7 @@ en:
     links: Links
     topics: Topics
   signs:
-    sign_of_the_day: Show me a sign
+    sign_of_the_day: Sign of the day
```
Br3nda, line 25: Is this a UI change?

config/routes.rb
```diff
@@ -49,5 +49,6 @@
   get '/assets/signs/:width-:height/*filename' => 'sign_image#show', :width => /\d+/, :height => /\d+/, :format => false
   get '/sitemap.xml' => 'sitemaps#index', defaults: { format: 'xml' }, as: :sitemap
 
+  get '/random_sign' => 'pages#random_sign'
```
Br3nda, line 52: maybe we should call it 'daily_sign' because it doesn't change enough to be a method to request for a random sign

spec/models/user_spec.rb
```diff
@@ -1,7 +0,0 @@
-# frozen_string_literal: true
-
-require 'rails_helper'
-
-RSpec.describe User, type: :model do
```
Br3nda, line 5: we should probably write some tests here though.

Br3nda (commented):
lots of comments. love your work

Br3nda:
@eoinkelly if you use phrases like `closes #576`, then the PRs are linked to the issue they close, and our waffle board will be tidy.

config/environments/development.rb
eoinkelly: This change is for local development only. Heroku envs will use memcached to store cache

config/locales/en.yml
eoinkelly: It is, I have provisional approval from @DSRU for it - this change will need to live in staging long enough for them to test it there.

config/routes.rb
eoinkelly: `/random_sign` is the destination when you hit the "Show me another sign" button on the homepage so it really is a random sign. The homepage will use the cached version but `/random_sign` will give you a new sign every time. @DSRU wants to keep the "Show me another sign" functionality and we still want the homepage to be fast/cached so I implemented it this way

eoinkelly:
> @eoinkelly if you use phrases like `closes #576`, then the PRs are linked to the issue they close, and our waffle board will be tidy.

@Br3nda This PR does not close #576 - #576 is further work to make the whole numbers page static - this PR just caches the numbers. I think #576 is desirable because it is simpler but there is enough work in it that I didn't want it in this PR

eoinkelly:
> @eoinkelly if you use phrases like `closes #576`, then the PRs are linked to the issue they close, and our waffle board will be tidy.

I hear you on this in general btw - I have been trying to remember to do it whenever relevant (apologies if I forgot) :smile:

Br3nda (approved):

config/locales/en.yml
DSRU: @eoinkelly how long do you suggest we keep it on staging?

eoinkelly: I wanted you to have time to test the feature - we can move it to prod if you are happy

DSRU: OK let's leave it at least a day then so that I can see how the caching works out.

Issue #576, opened 2019-02-25
Make the numbers page a static page


eoinkelly:
https://www.nzsl.nz/numbers currently pulls data from Freelex but the data never changes. @DSRU suggested we make this a static page and that seems sensible.

DSRU:
I have put all the necessary drawings for this page on Google Drive: https://drive.google.com/drive/folders/1cndFT0rP37ODOyqiRIvi97wp2144EdqP?usp=sharing

PR #584, opened 2019-02-28
Recover from unexpected cache errors

eoinkelly:
I tried and failed to replicate that cache error I saw on staging. Staging seems to work flawlessly during my testing today and I was never able to replicate the issue locally. To be safe I have created this PR which will recover from cache issues by re-trying the data fetch (without the cache). 

Rails is pretty good at dealing with situations where the cache goes away unexpectedly but just to be safe we are rescuing errors from cache lookups and recovering by fetching the data directly. We also log to Raygun so that these cache errors will be noticed if they are an issue.

### DIFF
--- a/app/helpers/numbers_helper.rb
+++ b/app/helpers/numbers_helper.rb
@@ -4,22 +4,30 @@ module NumbersHelper
   def numbers # rubocop:disable Metrics/AbcSize
     Rails.cache.fetch('numbers', expires_in: 7.days) do
       Rails.logger.info('Fetching new numbers signs from Freelex')
-
-      number_signs = {}
-      Sign.all(tag: 29).each { |s| number_signs[s.id.to_i] = s }
-
-      {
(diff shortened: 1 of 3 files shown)


Rabid-Dan (approved):
🌟 looks good once the code climate errors have been fixed

eoinkelly:
@Rabid-Dan  CodeClimate thinks I have duplication (https://codeclimate.com/github/ODNZSL/nzsl-online/pull/584) but I think that the code would be made worse by trying to pull those together 🤷‍♂️

Rabid-Dan:
@eoinkelly oh yep I agree, it does occasionally bring up annoying perceived duplication things like that. PR has a 👍 from me if you just want to ignore that "error"

eoinkelly:
BTW I didn't add tests to the `NumbersHelper` because we have issue #576 which will remove it entirely and make numbers be a static page

PR #590, opened 2019-03-12
Reduce time vocab sheets are cached

KatieWright26:
Per Mickey's request, I have reduced the number of days a vocab sheet is cached to 14 due to database limits.

https://rabidtech.atlassian.net/browse/CC-342


Rabid-Dan (approved):

elspeth-rabid (approved):
Remember to update the frequency the rake task is run on Heroku 💯

elspeth-rabid (commented):
This only changes how far back the rake task will delete vocab sheets from.
We need to update the lifetime of the cookie also, and update the frequency at which the rake task is run on Heroku.

KatieWright26:
@elspeth-rabid the rake task runs daily - did you want it running more often than that?

eoinkelly (approved):

PR #605, opened 2019-03-26
Bugfix/play icon not visible

l-suzuki:
This PR responds to: 
CC-407 https://rabidtech.atlassian.net/browse/CC-407 
Zendesk ticket #798 https://ackama.zendesk.com/agent/tickets/798 

The issue: NZSL had reports from Android users that the ‘play’ button has disappeared from the videos in both the search results and the sign pages when viewed on a mobile phone. The video still works when you click on it, but users don’t know to click on the video when there is no play button.

This only affects the responsive view of the website on Android (i.e. the full website view, Android mobile app and iOS mobile app are all fine)

There was a line in buttons.scss that was disabling the play button explicitly for the above use case. This has been removed. I also moved the logic pertaining to play icons in this file into _videos.scss as Elspeth and I agreed that this was more logical. 

In addition, I have refactored the logic for custom play buttons so that it responds to a screen resizing event listener as well as on page load. 

I have tested this fix using Browserstack on both android and iOs devices (new and old) without issues. Screenshot here showing android play icon displaying correctly: 

Screen Shot 2019-03-26 at 11 27 50 AM


l-suzuki:
@elspeth-rabid I tested this on Browserstack using iPhone X and it worked fine, so I am assuming the breakpoint 'large' is still ok because of this? But lmk if you think otherwise :)

app/assets/javascripts/custom_play_button.js
```diff
@@ -1,41 +1,59 @@
...
   });
 
-  $('.play-button').click(function(e) {
+  $(".play-button").click(function(e) {
     e.preventDefault();
     videoResponse(this.nextElementSibling);
   });
 
+  function adjustVideoControlsToScreenSize() {
+    if (Modernizr.touch && !Foundation.MediaQuery.atLeast("large")) {
```
elspeth-rabid, line 21: There needs to be an else to this situation where you show the play button and hide the controls.

app/assets/stylesheets/_buttons.scss
```diff
@@ -119,17 +128,3 @@ a {
   height: 1.5rem;
   margin-left: 0;
 }
-
```
elspeth-rabid, line 122: ✨

app/assets/stylesheets/_video.scss
```diff
@@ -35,3 +35,10 @@ video {
     display: block;
   }
 }
+
+// Hides the default videojs BigPlayButton because we have our own custom one
+.video-js {
```
elspeth-rabid, line 40: ✨

elspeth-rabid (changes requested):
Just the linting and the else case :-)

app/assets/javascripts/custom_play_button.js
l-suzuki: I'm not sure I understand why this is necessary - could you elaborate? :)

l-suzuki: As in -- isn't the else case you're describing the default situation? So if the 'if' conditions aren't met it just defaults back to that anyway?

PR #606, opened 2019-03-26
Bugfix/play icon not visible

l-suzuki:
This PR responds to:
CC-407 https://rabidtech.atlassian.net/browse/CC-407
Zendesk ticket #798 https://ackama.zendesk.com/agent/tickets/798

The issue: NZSL had reports from Android users that the ‘play’ button has disappeared from the videos in both the search results and the sign pages when viewed on a mobile phone. The video still works when you click on it, but users don’t know to click on the video when there is no play button.

This only affects the responsive view of the website on Android (i.e. the full website view, Android mobile app and iOS mobile app are all fine)

There was a line in buttons.scss that was disabling the play button explicitly for the above use case. This has been removed. I also moved the logic pertaining to play icons in this file into _videos.scss as Elspeth and I agreed that this was more logical.

In addition, I have refactored the logic for custom play buttons so that it responds to a screen resizing event listener as well as on page load.

I have tested this fix using Browserstack on both android and iOs devices (new and old) without issues. Screenshot here showing android play icon displaying correctly:

Screen Shot 2019-03-26 at 11 27 50 AM


elspeth-rabid (approved):
YES

PR #622, opened 2019-04-23
Release to prod

Br3nda:


Br3nda:
Contains a security patch, so i went ahead and merged it.

PR #636, opened 2019-05-09
Add visual regression testing

Br3nda:
Percy takes screen shots of capybara tests, and uploads them -- then compares them between branches on pull requests.

This will help us catch visual regressions.

Percy is one of the checks on the pull request.


Br3nda:
replaced with #637 
because the PERCY_TOKEN isn't available on forks

PR #701, opened 2019-07-17
Store VocabSheet items as JSON within VocabSheet instead of as separate table

### LINKED ISSUE #698: Store vocab sheet items inline in vocab_sheets table to stay within heroku DB plan row limits
Currently `VocabSheet` has many `Item` and during busy times the number of items stored can take us up to and over the 10k DB rows limit on our Heroku DB plan. Using a different plan isn't within the budget in the medium term so we should change items to be stored inline (possibly as a jsonb) in the VocabSheet model.

eoinkelly:
Change implementation of `Item` to not store as separate model.

* Start storing items within the `VocabSheet` as a JSON blob
* Convert `Item` from an `ActiveRecord` object to `ActiveModel` to preserve the public interface that the rest of the app expects.
* Add methods to `VocabSheet for handling Item storage and retrieval
* Remove unused `notes` field from Item

The motivation behind this change is staying under 10k DB rows on our current Heroku hosting plan because there is no extra funding for moving to a larger plan right now.

Minor bugs I fixed along the way:

* Fix JS error when changing name of vocab sheet
    * This didn't result in an error users would notice (the name change was saved and shown to them on the page) but it was causing an exception in the console
* Upgrade rubocop to 0.67 because old version could not parse Ruby 2.6 properly

This is a fairly substantial change so should get some manual testing in staging too @DSRU 

Closes #698


eoinkelly:
Travis can't run the migration for some reason, will look into it later.

Br3nda:
Add this to travis.yml

```
services:
  - postgresql
addons:
  postgresql: "10"
  apt:
    packages:
    - postgresql-10
    - postgresql-client-10
```

Br3nda:
(but also check what version we have on heroku)

app/models/item.rb
```diff
@@ -1,38 +1,31 @@
...
 
-  default_scope { order('position ASC', 'created_at ASC') }
+  attr_accessor :id,
+                :sign_id,
+                :drawing,
+                :name,
+                :maori_name
 
-  attr_writer :sign
+  def initialize(attrs = {}) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity
```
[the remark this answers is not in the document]
eoinkelly: I respectfully disagree here Mr. Bot.

eoinkelly:
Heroku is running Postgres 9.6.14 (which definitely supports JSONB columns). I have updated Travis to also use 9.6

Br3nda:
I'd like to give this a good test before merging. Can look on Friday

app/models/item.rb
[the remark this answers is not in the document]
Br3nda: I've approved on codeclimate. Also the bot is a ms.

eoinkelly:
Yup that seems like a good idea. I'm creating a review app on Heroku for it so we can all have a good click around

db/migrate/20190716173959_remove_items.rb
```diff
@@ -0,0 +1,15 @@
+class RemoveItems < ActiveRecord::Migration[5.2]
+  def change
```
Br3nda, line 2: Shouldn't we migrate existing vocab sheets here too?

elspeth-rabid (commented):
I am also reviewing this tomorrow @eoinkelly 📚

db/migrate/20190716173959_remove_items.rb
eoinkelly: I assumed we wouldn't bother with data migration given that vocab sheets are quite short lived but maybe that's being a bit 🤠 ?

Br3nda: i guess we don't know. We could do some analysis on the data to work out the usage patterns of people who use vocab sheets

Br3nda (approved):
works great

eoinkelly:
Closes issue #698  (forgot to mention this in original description)

Br3nda:
@eoinkelly merge this when you're ready.

app/models/vocab_sheet.rb
```diff
@@ -2,11 +2,93 @@
 
 ##
 # A "sheet" of items (signs) saved by a user
+#
 class VocabSheet < ApplicationRecord
-  has_many :items, dependent: :destroy
+  ##
+  # @param item [Item] The item you wish to add
+  # @return [Boolean] true on success, false on failure
+  #
```
elspeth-rabid, line 10: Are you a method documentation convert @eoinkelly ?

elspeth-rabid (approved):
Nice work and thanks @eoinkelly !

Br3nda:
@eoinkelly anything left here? Shall we merge this?

app/models/vocab_sheet.rb
eoinkelly: Kinda :-) I was trying to mimic the existing API that the controllers depended on so I needed to get the types straight in my head.

eoinkelly:
Yes I reckon we should.

eoinkelly:
I have a pretty full meetings day but I'll merge tonight when I have enough time to watch it go through.

DSRU:
Just tested on staging (which is running very slow on Chrome, a bit better on Firefox). All seems to be working except for broken links to the images when saved to pdf directly. If saving as pdf via the print menu, images download correctly. 
vocab sheet on staging

PR #724, opened 2019-08-09
Trial skylight.io application performance monitoring

eoinkelly:
Skylight (https://www.skylight.io) is a nice APM tool, similar in functionality to New Relic. I think it is worth trialling it **alongside** our existing New Relic because:

* We can see 24 hours of data with New Relic. Skylight will allow us to see a lot more history.
* The Skylight agent is very small and fast (written in Rust) so adds very little overhead to the application - my research indicates that it should be fine to run both it and New Relic
* Skylight is free for open source projects. If we find it useful then we can apply to be added to their "open source" plan - this would make our performance data public which I think would ultimately be a good thing because it will allow **all** open source contributors (not just those within Ackama) to find (and hopefully fix) performance issues. https://www.skylight.io/oss has links to some existing OSS projects which use Skylight - you can see the Skylight UI from there too.
* Skylight can be configured through environment variables so we can turn it on/off without redeploying - the `SKYLIGHT_ENABLED=true` env var must be set to enable Skylight - this will allow us to safely trial it in production.

If we find that Skylight covers all the key functionality of New Relic then we should remove the New Relic agent at that stage. I don't recommend us doing that until we have lived with Skylight for at least a few weeks.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,6 +51,9 @@ gem 'faraday_middleware'
 # logins
 gem 'devise'
 
+# Application performance monitoring
+gem 'skylight'
+
 group :assets do
   gem 'coffee-rails'
   gem 'sass-rails'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -316,6 +316,10 @@ GEM
       json (>= 1.8, < 3)
       simplecov-html (~> 0.10.0)
     simplecov-html (0.10.2)
+    skylight (4.1.2)
+      skylight-core (= 4.1.2)
+    skylight-core (4.1.2)
+      activesupport (>= 4.2.0)
     sprockets (3.7.2)
       concurrent-ruby (~> 1.0)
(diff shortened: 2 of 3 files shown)


Br3nda:
Are you using elsewhere? I've been using scout. This one looks better, but I'd be reassured if you're already using it.

eoinkelly:
We haven't used it anywhere but it's pretty commonly used in Ruby community. The company was created by Yehuda Katz so I trust it is competent. They are also quite focused on Rails which seems good. Bottom line is that we can turn off by flicking an env var if we don't like it for some reason so I'm keen to give it a go.

Br3nda (approved):

Br3nda:
@eoinkelly any privacy statement updates needed?

eoinkelly:
Good point @Br3nda, I don't think so - Skylight doesn't show anything which would identify a user so I don't think any change is needed.

PR #727, opened 2019-08-10
Skylight agent will not start with an empty config file

eoinkelly:
We are using environment variables to configure Skylight so had an empty config file. Skylight agent will not start with an empty config file.


eoinkelly:
@Br3nda I'm going to just merge this to staging now because it's trivial.

PR #730, opened 2019-08-17
Cache VocabSheet#items to mitigate performance regression

eoinkelly:
When we stopped implementing VocabSheet#items with an ActiveRecord relation
we stopped getting the benefit of the various layers of caching that AR
does for us. This caused a performance regression because

1. VocabSheet#items is called multiple times per request if the user has
  created a vocab sheet.
2. Every call to VocabSheet#items was hitting Freelex once for each item
   (So N items in a vocab sheet would result in N request to Freelex)

This change mitigates that performance regression by caching the items for
use later in generating the same response. It makes use of the ActiveRecord
`*_changed?` methods introduced in Rails 5.

I was able to track this down because Skylight showed we were making multiple requests to Freelex triggered by the views.  

I considered caching these item objects in memcache (via `Rails.cache`) but we only have 30MB of cache available and I think that items would probably exhaust that quickly.

### DIFF
--- a/app/models/vocab_sheet.rb
+++ b/app/models/vocab_sheet.rb
@@ -89,7 +89,7 @@ class VocabSheet < ApplicationRecord
 
     new_raw_items = []
 
-    item_ids.each do |id|
+    item_ids.uniq.each do |id|
       new_raw_items << raw_item_attrs.find { |item| item['id'] == id }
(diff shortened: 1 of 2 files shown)


Br3nda (approved):
tested, works nice

PR #732, opened 2019-08-17
Allow caching of Signs used for Item building in Rails cache

eoinkelly:
This change stores the Sign objects used to build vocab sheet Item objects in Rails cache. The objects are cached for a maximum of 24 hours.

This feature can be toggled on/off at runtime via an environment variables.

      FEATURE_CACHE_ITEMS_ENABLED=true # => enabled
      FEATURE_CACHE_ITEMS_NUM_HOURS=24 # default

The Sign objects used to build items are re-fetched on every request from a user who has a vocab sheet. This set of sign objects is much more constrained than the full set - the only Sign objects we will cache are those which are part of a vocab sheet so I predict that we can cache Sign objects in this case without overwhelming our small in-memory cache allocation.

I am using a feature flag because it is impossible to be certain about this prediction until we try it in production (with production load).

The only downside to exhausting our memcached allocation is that older items will get booted out sooner than the 24hr time limit I set in this PR. This change should provide a nice speed up to anybody loading pages when they have items in their vocab sheet.


Br3nda (commented):
Can we have a test? A spec that ensures it works both ways, and also checks we use the cache on the second call.

eoinkelly:
Added some tests. If this feature proves worthwhile I will make a future PR to remove the feature flag (and the associated tests). If it doesn't prove useful, I will also remove the feature flag because it'll be extra complexity we don't want.

eoinkelly:
I've had an idea for refactoring this into a better solution. Hold fire on reviews fo the moment

app/models/item.rb
```diff
@@ -29,4 +30,17 @@ def initialize(attrs = {}) # rubocop:disable Metrics/AbcSize, Metrics/Cyclomatic
   def to_param
     id
   end
+
+  private
+
+  def fetch_sign(sign_id)
+    if FeatureFlags::StoreVocabSheetItemsInRailsCache.enabled?
+      Rails.cache.fetch(sign_id, expires_in: FeatureFlags::StoreVocabSheetItemsInRailsCache.cache_timeout.hours) do
+        Rails.logger.info('Fetching sign from Freelex to build Item')
```
elspeth-rabid: Is this to be able to search the logs for analysing how often requests to freelex happen?
It might be nice to add the sign id but 100% 🐘

config/environments/test.rb
```diff
@@ -35,7 +35,7 @@
 
   # We want to test caching so give ourselves a little in-memory cache to use
   # for tests
-  config.cache_store = :memory_store, { size: 5.megabytes }
+  config.cache_store = :null_store
```
elspeth-rabid: ⭐️

config/initializers/feature_flags.rb
```diff
@@ -0,0 +1,29 @@
...
+      Integer(ENV.fetch('FEATURE_CACHE_ITEMS_NUM_HOURS', 24))
+    end
+
+    def self.status_msg
+      "#{self} is #{enabled? ? 'enabled' : 'not enabled'} (cache timeout: #{cache_timeout} hours)"
+    end
+  end
+end
+
+Rails.logger.info(FeatureFlags::StoreVocabSheetItemsInRailsCache.status_msg)
```
elspeth-rabid, line 29: I like this ^^^

spec/models/item_spec.rb
```diff
@@ -9,47 +9,104 @@
   let(:drawing) { 'foo.jpg' }
   let(:sign) { instance_double('Sign', id: sign_id, gloss_maori: maori_name, gloss_main: name, drawing: drawing) }
```
elspeth-rabid: I think we tend to use lowercase at the start of it/context blocks because it looks nicer when the test case sentence gets concatenated together.
🐘

elspeth-rabid (approved):
Cache it good

spec/models/item_spec.rb
eoinkelly: Good point, will change :sparkles:

eoinkelly:
This is still WIP. I think I have a good implementation now but the git history is a mess. Once I have tested it on the heroku app for this PR (hat tip @Br3nda for reminding me those are a thing), I will fixup the commit history and remove `[WIP]` from the title. Thanks for your patience y'all

eoinkelly:
This is ready to review on the Heroku app associated with this PR. If you go to the Memcached addon you can see graphs of number of object stored and number of MB used.

Running locally in verbose mode (`memcached -vv`), it seems that each sign takes up approx. 630 to 690 bytes in Memcached (from an admittedly small sample of signs). 

If we round up to 700 bytes/sign then we should in theory fit 1497 signs in 1 MB. Given that at the peak of sign language week we had approx. 10k vocab sheet items, I think we _should_ be able to comfortably fit the signs for the vocab sheet items in the cache.

PR #733, opened 2019-08-18
Fix one bug and performance improvements by refactoring

eoinkelly:
Moving object literals out of methods into constants mean they only have to be allocated once at Rails boot. Both `SignParser` and `SignMenu` are hit multiple times in rendering requests so there should be a minor performance improvement because of the refactoring.

### DIFF
--- a/app/models/sign_menu.rb
+++ b/app/models/sign_menu.rb
@@ -1,88 +1,143 @@
 # frozen_string_literal: true
 
-class SignMenu
-  def self.handshapes
+class SignMenu # rubocop:disable Metrics/ClassLength
+  HANDSHAPES = [
     [
-      [['1.1.1', '1.1.2', '1.1.3'], ['1.2.1', '1.2.2'], ['1.3.1', '1.3.2'], ['1.4.1']],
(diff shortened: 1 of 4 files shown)


app/views/vocab_sheets/show.html.haml
```diff
@@ -1,5 +1,3 @@
-- content_for :scripts do
```
Br3nda, line 1: Is this no longer needed?

eoinkelly: Yep. The only thing the block was injecting into the `:scripts` section was that typo'd comment

elspeth-rabid (approved):
Nice Eoin!

app/models/sign_menu.rb
```diff
@@ -1,88 +1,143 @@
...
+      ['6.5.1', '6.5.2'],
+      ['6.6.1', '6.6.2']
+    ],
+    [
+      ['7.1.1', '7.1.2', '7.1.3', '7.1.4'],
+      ['7.2.1'],
+      ['7.3.1', '7.3.2', '7.3.3'],
+      ['7.4.1', '7.4.2']
     ]
+  ].freeze
```
Br3nda, line 49: :sparkles:

Br3nda (approved):

PR #738, opened 2019-08-20
Make the numbers template static. Fixes issue #576

eoinkelly:
Fixes issue #576


app/views/pages/numbers.html.haml
```diff
@@ -1,26 +1,491 @@
...
+  .small-11.small-centered.medium-10.large-8
     .typography
       %h2#cardinal Cardinal Numbers
-    = render partial: 'static_pages/grid', locals: {group: numbers[:cardinal]}
-  - unless numbers[:ordinal].length.zero?
+    / with divs because of IE7
+    .typography
+      %ul.no-bullet
+        %li.small-6.medium-4.grid-stack.float-left.text-left
+          %a.drawing{:href => "#{sign_path("5655")}"}
```
Rabid-Dan: I guess this would be moving it back slightly in the original direction but could the sign path, image name, and alt for these get stored in a big ol hash and looped over instead?

app/views/pages/numbers.html.haml
```diff
@@ -1,26 +1,491 @@
-= render partial: 'pages/content'
-.small-11.small-centered.medium-10.large-8
-  - unless numbers[:cardinal].length.zero?
+- cache "numbers-fragment" do
+  .typography.small-11.small-centered.medium-10.large-8
+    %h1 Numbers
+    .clearfix_left.column-block
+      .translation_video{"data-ratio" => "0.75", :style => "display: none;"}
+        %video.video-js.vjs-default-skin{:controls => "controls", :controlslist => "nodownload", :loop => "loop", :preload => "auto"}
+          %source{:src => "http://s3-ap-southeast-2.amazonaws.com/nzsl-assets/number-signs.mp4", :type => "video/mp4"}
```
Br3nda: How about using the newer ruby hash syntax?

Br3nda (approved):
:elephant:

app/views/pages/numbers.html.haml
eoinkelly: I created this version of the file backwards from the HTML in browser which is why everything is unrolled. This page won't change at all until (if) we do a re-design so I'm inclined to be lazy and leave it as it is.

eoinkelly: That said, I'm happy to change it if y'all want me to.

Br3nda:
@eoinkelly do you need me to merge this one?

PR #739, opened 2019-08-21
Production deploy

eoinkelly:
Staging now has caching of vocab sheet signs enabled and it seems to be working well. 

I encourage y'all to have a click around in case I've missed anything. Having this hit production should mean that the site will be just as fast for people with vocab sheets as those without.


DSRU:
Fix website footer text being saved as part of PDF file, and signs overlapping with copyright footer on pdf, before putting into production.
vocab_sheet_august_footer

Br3nda (approved):

eoinkelly:
PDF issue fixed now. Got :thumbsup: from @DSRU on Slack so merging now.

PR #746, opened 2019-08-29
Update ruby version in Gemfile.lock

eoinkelly:
This change was made by running `bundle install` on the new Ruby version.
Heroku deploys fail if Gemfile and Gemfile.lock don't match.


eoinkelly:
@Br3nda I'm going to just merge this one because it's 1) tiny and 2) blocking deploys.

PR #747, opened 2019-08-29
Fix bug in VocabSheetsController#download_pdf

eoinkelly:
VocabSheetsController#download_pdf was refactored out of
VocabSheetsController#show. During the extraction I didn't realise that
a number of templates test the value of `params[:print]` directly and
make choices based on its value. This change sets the value as the
templates expect it.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -424,7 +424,7 @@ DEPENDENCIES
   will_paginate
 
 RUBY VERSION
-   ruby 2.6.3p62
+   ruby 2.6.4p104
 
 BUNDLED WITH
    1.17.3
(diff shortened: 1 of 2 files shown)


eoinkelly:
Urgh sorry everybody I accidentally pushed this directly to staging. Given how small this is, I'm not going to undo that because I'm assuming you are all :thumbsup: on this change. If not I'll do the necessary git faffing to make it right. Until then, this PR is so you can see the change.

eoinkelly:
The commit that I pushed was https://github.com/ODNZSL/nzsl-online/commit/1b262ae8c2001a31d995092c3623acb229dfde96

PR #749, opened 2019-08-30
Refactor Sign model

eoinkelly:
I'm exploring caching all sign searches locally in memcached. As part of that I needed to understand the `Sign` API so I did some gardening while I was there. This PR does not change **any** functionality in the app - it is just a refactor.

* Add yardoc comments to indicate param and return types to help me
  understand the Sign public API
* Use `class << self` syntax to allow us to have both public and private
  class methods.
* Remove some unused methods from Sign
* Refactor StaticPagesHelper to allocate less and remove unused method


app/models/sign.rb
```diff
@@ -51,193 +52,218 @@ class FreelexCommunicationError < StandardError; end
...
-  def self.from_json(json)
-    return nil if json.nil? || json.empty? || json == 'null'
+      SignParser.new(xml_nodeset.first).build_sign
+    end
 
-    sign = Sign.new
+    # @param search_query_params [Hash] The search query
+    # @return [Array<Sign>]
+    def all(search_query_params)
+      all_with_count(search_query_params)[1]
```
Br3nda, line 105: coverage is saying we don't call this method in the test anywhere

app/helpers/static_pages_helper.rb
```diff
@@ -1,13 +1,12 @@
...
+    A B_open B_closed C_half C_full D E F G H I J K L
+    M N O P_open P_closed Q R S T U V W X Y Z
+  ).freeze
 
-  def link_sign(name, id)
-    return link_to name, sign_path(id) if Sign.find(id: id)
-
-    name
+  def alphabet_listing
+    ALPHABET_LISTING
```
Br3nda, line 10: is this still needed? I think you've refactored out all the calls to it now

eoinkelly:
The changes here are included in PR #750 so declining this.

PR #750, opened 2019-08-30
Simplify the Sign model (chains on PR #749)

eoinkelly:
This PR chains on PR #749. I have split them like this to make them easier to review.

### Simplify the Sign model

This change does not change the public interface of Sign.

I have refactored the internals of Sign as follows:

I observed that `Sign.first` is always either called with a sign ID `{id: "1234"}` _or_ `{random: 1}`. I tested this observation by logging the params to `Sign.first` and running the test suite.

This means that the searches that `Sign.first` does are always going to return either 0 or 1 signs. This in turn means that we can make `Sign.first` use `Sign.all` for fetching signs without a performance penalty because we know we won't end up building a bunch of signs objects in `Sign.all` and then discarding all but the first.

The above allowed me to simplify the code flow through `Sign` as shown in the following diagram:

```
Sign.all -> Sign.search -> ...
Sign.first -> Sign.all -> Sign.search -> ...
Sign.paginate -> Sign.all -> Sign.search -> ...
Sign.random -> Sign.first -> Sign.all -> Sign.search -> ...
Sign.find_by_id_via_cache -> Sign.first -> Sign.all -> Sign.search -> ...
Sign.sign_of_the_day -> Sign.random -> Sign.first -> Sign.all -> Sign.search -> ...
```

This refactoring makes the `Sign` object easier to work with and sets up `Sign.all` as a natural place for any later caching we may wish to add.


app/models/sign.rb
```diff
@@ -51,193 +52,203 @@ class FreelexCommunicationError < StandardError; end
...
-    # value of the block to the cache
-    #
-    cached_sign_json = Rails.cache.read(SIGN_OF_THE_DAY_CACHE_KEY)
+    # Creates a new instance of a `Sign` object given a set of JSON attributes.
+    # This method is designed to be used as the inverse of `Sign#to_json`
+    # @param json [String] JSON representation of a `Sign` as created by `Sign#to_json`
+    # @return [Sign] if we successfully found a sign
+    # @return [nil] if we did not find a sign
+    def from_json(json)
+      return nil if json.nil? || json.empty? || json == 'null'
```
Br3nda, line 158: When is nill ever passed in? What's the scenario?

eoinkelly: In theory never but we are getting the string that represents the sign over the network (from Freelex) so I thought it was sensible to cover as many of the normal error cases

Br3nda (approved):

PR #751, opened 2019-08-31
Cache all signs [WIP, HOLD]

eoinkelly:
_This PR will be much easier to review once PR #750 is merged_

Cache all Sign objects in `Rails.cache`

This change causes all Signs from Freelex to be cached in `Rails.cache` as JSON.

The cache key is the query string sent to the Freelex because this string is unique for each set of search results returned.

The signs are cached for a configurable number of hours (default is 24). The cache expiry is configurable at runtime because we have a limited amount of cache space available and expiry will have to be tuned.

The caching is under a feature flag at runtime so we can stop if we need to.


eoinkelly:
There is a Heroku review app available for this PR now (for those with Heroku access) which has caching of signs enabled.

eoinkelly:
Currently this PR caches **search results** (a search-term and an array of signs). I did this because in order to avoid the round-trip to Freelex we would need to have the answer for a full search available. 

The downside of this approach is that signs may be cached multiple times e.g. searches for _head_ and _headmaster_ are different searches so their results would be cached separately but they would have signs in common.

This uses more memcache memory in exchange for the app having to do only one cache lookup to find the results in the cache. 

There is another approach, where I cache the search as  `(search-term, array-of-resulting-sign-ids)` and then store each sign individually in the cache. This would be make more efficient use of cache memory at the cost of the application having to do N+1 cache reads to get a result of N signs from the cache. Cache reads are pretty quick so this would still be _much_ faster than going to Freelex for the result.

After thinking about this for a while I'm going to update this PR to use the latter approach. We can always roll back in git history if we decide to go with the current approach.

Br3nda:
This is a fabulous idea Eoin.

Br3nda:
i've removed my name from the reviewers. Can you add again once you've finished and not WIP

Br3nda:
I think we could do some smart things here, by saving the signs into elastic search

PR #769, opened 2019-09-23
Bugfix/missing handshape images

eoinkelly:
Fix bug preventing handshapes from being visible in "Advanced search"

This bug was caused by us calling Array#shift on the handshapes. This bug was
introduced by the performance optimisation which move those into constants from
being re-generated each time.


louise-r-blue (approved):
Cool change and approve but why do constants need `.freeze` when they are constant and unchangeable?

Issue #775, opened 2019-09-30
Topic searches are not paginating properly


eoinkelly:
Only the first page of each topic is showing e.g. https://www.nzsl.nz/signs/search?tag=5

eoinkelly:
I'm looking into this now. So far, I think the issue is on our end not Freelex.

PR #776, opened 2019-09-30
Fix pagination bug

eoinkelly:
Fixes issue #775 

Fix bug preventing pagination of search results

* Fix bug in `Sign.paginate` where the count of total results was omitted preventing pagination links from displaying.
* Add a test verifying that pagination works


eoinkelly:
This bug was introduced by me refactoring `Sign` a while back. Seems the test suite isn't all it could be. I have added a test for this particular bug.

Rabid-Dan (approved):

PR #858, opened 2020-01-05
Clear the href of the video wrapper

### LINKED ISSUE #378: NZSL- 20. Bug on live, staging and dev. An error occurs when clicking the video play/pause button. Displays an error in Safari
Clicking the pause/play button on Demonstration video on the https://nzsl-staging.herokuapp.com/alphabet/ causes an error. 
May be a Safari specific bug.
One screenshot shows the web inspector out-put.
The link wrapped around the video has a href of "alphabet-wide" when it should be "#alphabet-wide".
The link wrapped around the video has a href of "alphabet-over-shoulder" when it should be "#alphabet-over-shoulder".
video_stop_pause-error_in_safari
safari_video-error

simonwelsh:
Safari on macOS will still treat the link as being clicked when using video controls, so remove the destination href so that users at left on the page.

Fixes #378


Br3nda:
Nice. I've tested and it still work in Chrome and Firefox.

Thanks for your contribution!

Br3nda (approved):

PR #914, opened 2020-03-18
Auto corrected by following Lint Ruby MagicComment

Br3nda:
Auto corrected by following Lint Ruby MagicComment

Click [here](https://awesomecode.io//repos/ODNZSL/nzsl-online/config_groups/ruby/188) to configure it on awesomecode.io


eoinkelly (approved):

eoinkelly:
CodeFactor wants me to use unary + to make a mutable string - that is pretty opaque syntax so my view is CodeFactor is wrong here. I can't resolve it in CodeFactor :shrug:

eoinkelly:
@Br3nda CodeFactor seems to be running Rubocop under the hood (a different version with different settings to what codeclimate runs). I think we need to have just one version of rubocop running or we'll have a lot of busy work to do - do we need both CodeFactor **and** CodeClimate?

Br3nda:
Agreed. Let's turn off code factor

eoinkelly:
Sweet. I have removed the Codefactor webhook.

PR #938, opened 2020-04-07
Bundle update

Br3nda:
attempting to fix bundle errors on travis-ci


eoinkelly:
Hey I think this update might be a bit broad - I think https://github.com/ODNZSL/nzsl-online/pull/943 is enough to fix this.

PR #949, opened 2020-04-13
Auto corrected by following Lint Ruby Lint/SendWithMixinArgument

Br3nda:
Auto corrected by following Lint Ruby Lint/SendWithMixinArgument

Click [here](https://awesomecode.io//repos/ODNZSL/nzsl-online/lint_configs/ruby/10173) to configure it on awesomecode.io

### DIFF
--- a/config/initializers/strong_parameters.rb
+++ b/config/initializers/strong_parameters.rb
@@ -1,3 +1,3 @@
 # frozen_string_literal: true
 
-ActiveRecord::Base.send(:include, ActiveModel::ForbiddenAttributesProtection)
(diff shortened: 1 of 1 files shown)


config/initializers/strong_parameters.rb
```diff
@@ -1,3 +1,3 @@
 # frozen_string_literal: true
 
-ActiveRecord::Base.send(:include, ActiveModel::ForbiddenAttributesProtection)
```
Br3nda, line 3: @eoinkelly Any idea why this was like this????

Br3nda: i guess this is an old old old app and that was the Rail 2 way..

PR #957, opened 2020-04-16
Ruby upgrade to fix secuity vuln

eoinkelly:


eoinkelly:
Merging now because I have time to respond if it goes wrong now.

PR #961, opened 2020-04-16
Auto corrected by following Lint Ruby Rails/DynamicFindBy

Br3nda:
Auto corrected by following Lint Ruby Rails/DynamicFindBy

Click [here](https://awesomecode.io/repos/ODNZSL/nzsl-online/lint_configs/ruby/25106) to configure it on awesomecode.io


spec/models/sign_spec.rb
```diff
@@ -113,8 +113,8 @@
         it 'does not cache calls to Freelex when creating the same Sign repeatedly' do
           expect(Sign).to receive(:first).twice
 
-          Sign.find_by_id_via_cache(sign_id)
-          Sign.find_by_id_via_cache(sign_id)
+          Sign.find_by(id_via_cache: sign_id)
```
Br3nda, line 116: nope. I think we could rename that method to avoid confusion.

eoinkelly:
Everything in here is covered by PR #982 so I think we can close this.

PR #977, opened 2020-04-26
Faster front end

eoinkelly:
This change includes a number of performance improvements. The goal is to

1. Make the pages load more quickly
1. Download the minimum amount of data to users. If users are visiting from mobile connections they pay by the MB so we want to be respectful of that.

#### JS changes

* Split JS into a bundle for the admin pages and public pages. It turns out we were loading a lot of JS for the public pages which was only required for the (logged in) admin pages. 
* Stop loading all of JQueryUI - we are now pulling in just the parts we need.

These JS changes make the JS downloaded to users much smaller - it drops from 396kB -> 96kB (compressed) which is 1.4MB -> 332kB uncompressed.

#### Font changes: 

* Allow the default fonts to display the page if _Montserrat_ font from Google fonts is loading slowly - this will cause a layout change when the real font comes in but that seems less bad than showing a white screen.

#### Video changes

* Changed `rel="auto"` to `rel="metadata"` - this stops the browser from downloading all the videos on the page in the background. A video will now be downloaded when the user clicks on it.

I have created a staging site at https://nzsl-faster-front-end-xocta7hz.herokuapp.com/ for this PR (the creds are the same as the usual staging site) so y'all can do manual testing as required.


eoinkelly:
The changes to the video I made cause a slight movement of elements on the page so Percy flagged this change. I manually inspected the change and it seems fine. I'll wait for y'all to review the staging site before I merge because that's the most meaningful test.

Br3nda:
Looks like you've also fixed the video controls regression. good work

app/assets/javascripts/admin_application.js
```diff
@@ -0,0 +1,30 @@
+/* eslint spaced-comment: "off" */
+
+// This JS is only loaded on Admin pages.
+
+//= require jquery
```
Br3nda, line 5: it seems odd to me that jquery is only loaded by the admin pages...

Br3nda: maybe a bunch of these can be removed, as they're already loaded in application.js

app/views/signs/show.html.haml
```diff
@@ -47,9 +47,9 @@
           .typography.videos
             .video-container
               %i.fi-play.play-button
-              = videojs_rails sources: { mp4: example[:video]}, class: "example_video video normal", controls: true, controlsList:"nodownload noremoteplayback", preload: "auto"
+              = videojs_rails sources: { mp4: example[:video]}, class: "example_video video normal", controls: true, controlsList:"nodownload noremoteplayback", preload: "metadata"
               - if example[:video_slow]
-                = videojs_rails sources: { mp4: example[:video_slow] }, class: "example_video video slow", controls: true, controlsList:"nodownload noremoteplayback", preload: "auto", width: "100%"
+                = videojs_rails sources: { mp4: example[:video_slow] }, class: "example_video video slow", controls: true, controlsList:"nodownload noremoteplayback", preload: "metadata", width: "100%"
```
Br3nda, line 52: :+1:

app/assets/javascripts/admin_application.js
eoinkelly: `jquery` is still loaded for the public facing pages, just not by this file.

The public facing part of the site is rendered by `app/views/layouts/application.html.haml` which loads `app/assets/javascripts/application.js` which in turn loads jquery.

The admin part of the site is rendered by `app/views/layouts/admin.html.haml` which loads `app/assets/javascripts/admin_application.js` which is this file.

Rabid-Dan (approved):
I dont have access to staging creds so I cant do any manual testing on the test site you've spun up but this looks like a really good set of changes to me.

app/assets/javascripts/admin_application.js
Br3nda: I usually see this done by including application.js in admin.js.

eoinkelly: In this case I think that a lot of the front-end JS isn't actually required for the admin but we are including it. I didn't bother optimizing it because basically nobody uses the admin anymore and I was running out of time.

Br3nda: :elephant:

Br3nda (approved):

Issue #978, opened 2020-04-26
Sanitize search queries sent to Freelex


eoinkelly:
I believe that

> the only other legitimate characters apart from letters and numbers would be a hyphen and a space. 

so we can do some filtering to help reduce the volume of malicious requests that get to Freelex.

PR #979, opened 2020-04-26
Sanitize queries we send to Freelex

eoinkelly:
Fixes issue #978 

This implements sanitization of **all** the query parameters we receive before sending them to Freelex. The rules for sanitization vary by parameter but hopefully the logic can be understood from the code.

We should test this manually when it hits staging to be sure that it hasn't changed any of the queries.


eoinkelly:
Despite Codeclimate's feelings on the matter, I stand by my layout of `sanitize_for_standard_search` - I think breaking it up would actually make it harder to discern the pattern in the (fairly complex) business logic

app/services/search_query_sanitization_service.rb
```diff
@@ -0,0 +1,122 @@
...
+
+  ##
+  # @param [String] handshape
+  # @return [String]
+  #
+  def sanitize_handshape(handshape)
+    return '' if handshape.nil?
+
+    handshape
+      .gsub(ANYTHING_EXCEPT_NUMBERS_SPACE_PERIOD_REGEX, '')
```
Br3nda, line 66: the long names did make me smile, but they are good names

Br3nda: (but btw in NZ we call that a fullstop, not a period, afaik)

eoinkelly: Same in Ireland but that was even longer! 😀

PR #982, opened 2020-05-03
Make the whole codebase pass rubocop

eoinkelly:
This change makes the whole codebase pass rubocop. This is desirable because:

1. It means that editors can be configured to invoke rubocop and give real-time feedback
1. It means we can run Rubocop in CI and fail if it fails. This will allow us to disable rubocop in all the other bots that run checks after commit. Having just one thing running Rubocop which can also be run locally will help a lot with getting good quality PRs before they are put up for review - currently we only find out about rubocop issues from the bots after the fact and sometimes the bots disagree because they are using slightly different versions of rubocop.

This PR achieves a passing rubocop by

1. Loosening some restrictions e.g. method length. I don't think any of the looser values are terrible. We could tighten them up later on if we wanted to.
1. Fixing a bunch of minor issues e.g. missing uniqueness validations
1. Disabling rules on specific lines when it would be too much work to refactor the old usage away. These disabled rules can be easily found in the codebase (by grepping for `rubocop:disable` so they can be addressed individually later on if somebody is keen to do that.


config/initializers/devise.rb
```diff
@@ -3,7 +3,7 @@
 # Use this hook to configure devise mailer, warden hooks and so forth.
 # Many of these configuration options can be set straight in your model.
 Devise.setup do |config|
-  config.secret_key = 'dabd54db3b77f2b880cffc1e90c85270063db8c636eb3280dd0c81bf40f633c3901a5c3d0ec04cb5965c2ab9aceac8e39ea08ad6b7dd3f431d2b4651080e586b'
+  config.secret_key = ENV.fetch('DEVISE_SECRET_KEY')
```
eoinkelly, line 6: I have already generated new values for this secret (by running `bundle exec rails secret` multiple times) and put those values into the env for staging and prod. When the new secrets are activated after restart then any existing password resets in the DB will need to be triggered again. Given that AFAIK **nobody** logs in to the admin interface these days this shouldn't disrupt anything.

eoinkelly:
Travis is failing because it needs `DEVISE_SECRET_KEY` added to it's env. I'm not a repo owner so I can't do this. @Br3nda can you update it when you get a chance please? Alternatively maybe I should be a repo owner?

app/helpers/application_helper.rb
```diff
@@ -14,18 +14,18 @@ def render_navigation_link(link)
   def submit_button(text = 'search.submit', options = {})
     "<div class='button input_button'>
```
Br3nda: could this be moved into a view instead of invoking htmlsafe?

app/helpers/signs_helper.rb
```diff
@@ -17,7 +17,7 @@ def render_grammar_notes(sign)
                 'Sign', 'Click', 'glossary #{note}']);" }
       link_to(t("signs.show.field.#{note}"),
               "#{Page.find(Setting.get(:glossary)).try(:path)}##{note}", attrs)
-    end.compact.join(', ').html_safe
+    end.compact.join(', ').html_safe # rubocop:disable Rails/OutputSafety
```
Br3nda: gosh, there are a lot of these.

I would rather leave them as failing cops, then they appear on a list to be fixed, instead of making them to pass silently.

Br3nda:
I'm adding you to the repo, so you should be able to edit travis in the future

Br3nda:
@eoinkelly you are already the owner of this repo - there's no more privileges i can give you. Don't know why you can't edit Travis-ci config

Br3nda:
DEVISE_SECRET_KEY is now in travis-ci

I made it a public value, so it will be available to any pull requests that come from a fork of this repo too.

app/helpers/application_helper.rb
eoinkelly: It could but that blows this story out quite a bit ...

eoinkelly:
I had a go at removing `html_safe` from at least the ApplicationHelper. Those helpers are quite a tangle so it's very slow going

Br3nda (approved):

app/helpers/application_helper.rb
```diff
@@ -2,72 +2,49 @@
 
 module ApplicationHelper
   def page_title
-    "#{@title}#{' -' if @title} #{t('layout.title')}"
+    "#{@title}#{' -' if @title} #{t('layout.title')}" # rubocop:disable Rails/HelperInstanceVariable
```
Br3nda, line 5: would really rather leave these as failing cops to be fixed. but i've hit approve anyways

eoinkelly:
I have refactored away all disabling of the `Rails/OutputSafety` cop. The full list of disabled cops in the codebase is now:

```
[17:05:27] ❯ ag "rubocop:disable"
app/helpers/search_helper.rb
3:module SearchHelper # rubocop:disable Metrics/ModuleLength
44:    return if @query[:hs].blank? # rubocop:disable Rails/HelperInstanceVariable
46:    query_hs = @query[:hs] # rubocop:disable Rails/HelperInstanceVariable
49:    query_hs = @query[:hs].map { |q| "#{q}.1" } if shape.split('.').last == '1' # rubocop:disable Rails/HelperInstanceVariable
55:    'selected' if @query[:l].present? && @query[:l].include?(location.split('.')[1]) # rubocop:disable Rails/HelperInstanceVariable
59:    'selected' if @query[:lg].present? && @query[:lg].include?(location_group.split('.')[0]) # rubocop:disable Rails/HelperInstanceVariable
85:    keys = @query.select { |_key, value| value.present? }.keys # rubocop:disable Rails/HelperInstanceVariable
102:    return if @query[:l].blank? # rubocop:disable Rails/HelperInstanceVariable
120:    return if @query[:hs].blank? # rubocop:disable Rails/HelperInstanceVariable
137:    return if @query[:lg].blank? # rubocop:disable Rails/HelperInstanceVariable
153:    h SignMenu.usage_tags.select { |u| @query[:usage].include?(u.last.to_s) }.map(&:first).join(' ') if @query[:usage].present? # rubocop:disable Rails/HelperInstanceVariable
157:    h SignMenu.topic_tags.select { |u| @query[:tag].include?(u.last.to_s) }.map(&:first).join(' ') if @query[:tag].present? # rubocop:disable Rails/HelperInstanceVariable
161:    return if @query[key].blank? || (@query[key].is_a?(Array) && @query[key].reject(&:blank?).blank?) # rubocop:disable Rails/HelperInstanceVariable
163:    h @query[key].join(' ') # rubocop:disable Rails/HelperInstanceVariable
176:    @display_search_term ||= safe_join(search_term_elements, ' ') # rubocop:disable Rails/HelperInstanceVariable

app/helpers/application_helper.rb
5:    "#{@title}#{' -' if @title} #{t('layout.title')}" # rubocop:disable Rails/HelperInstanceVariable

app/helpers/vocab_sheet_helper.rb
5:    return nil if @sheet.blank? # rubocop:disable Rails/HelperInstanceVariable
6:    return nil if @sheet.items.length.zero? # rubocop:disable Rails/HelperInstanceVariable
13:    return offsets[:"size_#{@size}"] if @size.present? # rubocop:disable Rails/HelperInstanceVariable

app/services/autocomplete_search_service.rb
31:    # rubocop:disable Style/AsciiComments

app/helpers/classifiers_helper.rb
4:  def classifiers # rubocop:disable Metrics/AbcSize, Metrics/MethodLength

app/services/search_query_sanitization_service.rb
17:  def sanitize_for_standard_search(params) # rubocop:disable Metrics/AbcSize

config/initializers/static_image_setting.rb
3:SIGN_IMAGE_PATH = "#{Rails.root}/public/system/images/signs/" # rubocop:disable Rails/FilePath

spec/services/temp_dir_janitor_service_spec.rb
22:          FileUtils.touch(removable_file, mtime: 1.1.hours.ago.to_time) # rubocop:disable Rails/Date
23:          FileUtils.touch(keepable_file, mtime: 0.9.hours.ago.to_time) # rubocop:disable Rails/Date
```

Of these, I think the next ones I would target are in `app/helpers/search_helper.rb` - I have created issue #1004 to track this.

The other instances of us disabling cops seem ok to me (I mean, yeah they would because I disabled them I guess :-) )

app/helpers/application_helper.rb
eoinkelly: Yeah I get that. Issue #1004 tracks fixing these but I think there's enough benefit to having the whole codebase pass that it's worth merging.

PR #983, opened 2020-05-03
Remove deprecated browser detection

eoinkelly:
Most of the browser detection that was happening no longer applies - the
only detection that is still relevant is checking for IE 9 and older.

This change does this with a conditional comment instead of trying to
do browser detection in Rails.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,10 +4,8 @@ class ApplicationController < ActionController::Base
   include BasicAuthHelper
 
   protect_from_forgery with: :exception
-  require 'browser'
   layout :layout_by_resource
 
-  before_action :check_browser_support
   before_action :staging_http_auth
 
   private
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -20,14 +18,6 @@ def layout_by_resource
     devise_controller? ? 'admin' : 'application'
   end
 
-  def setup_browser_rules
-    Browser.modern_rules.clear
-    Browser.modern_rules << ->(b) { b.chrome? && b.version.to_i >= 55 }
-    Browser.modern_rules << ->(b) { b.firefox? && b.version.to_i >= 51 }
-    Browser.modern_rules << ->(b) { b.safari? && b.version.to_i >= 9 }
-    Browser.modern_rules << ->(b) { b.ie? && b.version.to_i >= 10 }
```
eoinkelly, line 28: Current versions of everygreen browsers

* Chrome = 80
* Firefox = 75
* Safari = 13

so I think we can safely ignore checking any of those. We still need to display a message to IE 9 or older - see the conditional HTML comment below

Br3nda: Hey  - just checking you ran this past Micky? She had some good reasons to support older browser for some Deaf users and their families who have older browsers.

eoinkelly: This change doesn't actually change our browser support at all. It just changes **how** we present the "your browser is not supported" message.

Br3nda (approved):

eoinkelly:
@DSRU FYI The site doesn't currently work on IE9. It works fine on IE10 and above. I'm mentioning this in case this is news to you. This change doesn't change that in any way.

app/controllers/application_controller.rb
DSRU: We agreed not to support IE legacy browsers - our Analytics showed that it was a very small user group in comparison with Chrome and Safari.

PR #1031, opened 2020-06-09
Production deploy

eoinkelly:
There is a websocket-extensions security fix in this that should be deployed


DSRU:
Everything looks fine on staging, but as usual I'm not sure what I'm looking out for.

I have noticed a few alignment issues (already on production though, so not necessarily to do with the current deploy). Firstly, the example sentence videos on the sign pages should be left aligned. Secondly, the 'show another sign' button should sit next to the 'sign of the day' video on the home page (rather than below it).


________________________________

Br3nda (approved):

PR #1042, opened 2020-06-20
Replace Travis with Github actions

eoinkelly:
Github actions are more convenient because they are fully integrated with Github and so permissions on the repo automatically become permissions on the CI. I still can't access the settings in Travis - I know I _should_ be able to (as a repo owner) but I can't so :shrug:


G-Rath:
Since this is a public OSS repo, action minutes are free and uncapped, so you could breakout the actual "actions" (i.e tests, RoboCop, and eslint) steps into their own jobs, so that they run in parallel.

This will result in some duplicate steps in the workspace, but should greatly reduce the ci feedback time :)

(The workspaces for eslint-config-ackama & eslint-plugin-jest are good examples of this)

Br3nda:
Nice work

Br3nda (approved):

eoinkelly:
@G-Rath Ooh good tip. The test suite here is only 7m total so for now I'll keep it with the simpler version but that's good to know for future

Br3nda:
i've turned off the requirement that travis passes on a PR

PR #1061, opened 2020-06-29
CCSD-1297/ensure search does not remove macron at start

G-Rath:
We got a ticket about inconsistent results for searches, which the reporter suspects "it only affects words that begin with a capital 'Ō' with a macron."

While we have a test already for checking we don't strip macrons, it's not testing that we don't when the macron is at the start; while in theory there shouldn't be a reason for the starting letter to be treated differently, I figure might as well just chuck a test case in anyway.


G-Rath:
Did this from the fork by mistake 😬

PR #1074, opened 2020-07-10
Production release

eoinkelly:
Includes Creative commons link update


G-Rath:
I've made #1075 to patch the latest known advisories - would be cool if we could get that merged for deployed here too :)

PR #1097, opened 2020-07-28
Add functionality to handle Freelex being down or unavailable

joshmcarthur:
Background:

The Freelex service frequently becomes unavailable. This is usually because VUWs firewall seems to be quite dependent on static IPs, and tends to blacklist our Heroku dyno every now and then. When this happens, we don't really have recourse other than to wait for Victoria University's IT staff to get to the support ticket. 

@eoinkelly and I were exploring the idea of somehow serving the sign data from our own database, and moving the Freelex dependency further back in the application to the build stage. This would mean that we did not have a runtime dependency on Freelex being available to us (other than the image caching - more on that later). 

I raised the point that the native iOS and Android applications both already to this, to support working offline, with this working well a number of years. We have the [nzsl-dictionary-scripts](https://github.com/odnzsl/nzsl-dictionary-scripts) repository, which fetches, parses and transforms a Freelex XML data dump into a TSV file (nzsl.dat), and a SQLite3 database (nzsl.db), which are then used by the Android and iOS applications respectively.

This pull request is an MVP solution to allowing the site to retain limited functionality if Freelex is unavailable. I don't see this as the final iteration on this problem, but I think it might be a useful addition to find out the edge cases of taking this approach. 

This solution takes the unmodified SQLite database from the iOS app, and establishes a connection from a model called `OfflineSign` to it. `OfflineSign` has a few column aliases and other tweaks to allow the `signs/show` template to render this model, with a reduced, but still useful set of information:

image

Rails 6 does support [multiple database connections](https://edgeguides.rubyonrails.org/active_record_multiple_databases.html), but this didn't appear to support the kind of mixed-tenancy model I was after, with data from one table living in one database, and data from another table living in another. The Rails 6 connection features appeared to be focused on scaling options - horizontal sharding, failover, and write/read roles. This could certainly support our use case, but it seemed less magic-prone to me to just `establish_connection` to the SQLite database in the model that needed it. This would be something I would no longer recommend if we ever scaled this to multiple models.

It also introduces feature flags to a) control whether Freelex is used at all for searches and sign look up (if Freelex is disabled, we present a banner to let the user know that some features are unavailable), and b) control the timeout to Freelex. The timeout is incidentally configurable - it was most added so that if Freelex was disabled, it could default to zero rather than twenty seconds. Having it be configurable allows for some flex though - for example, if see that Freelex is up, but there is known maintenance or issues, we can shorten the timeout so that requests will fail quickly rather than leaving the user hanging and then failing.

One feature that has had to be bypassed is the image processing and caching. Instead, the image associated with the sign must be hosted somewhere publicly accessible. In my testing, I was using `public/images/signs`, but this adds many files to the git repository, and I didn't want to check these in since they would then be in the history forever. 

The reason we bypass image processing is that image processing relies on the Heroku app being able to contact the nzsl-assets.vuw.ac.nz domain to fetch the image server-side. This means that if the firewall has kicked in, the sign data will be fetchable, but the image would still fail to load. By returning a direct URL for the sign picture to the client, we can work around the typical IP block in the firewall, since it's the user's browser that fetches the image, not our server.

I have deliberately used the SQLite database from the native apps for this MVP, though it requires more processing of the data. The apps only include the columns they require, so things like IDs, examples, usage notes, and the actual URLs to pictures and videos are not in this database. Because of this, there's a certain amount of processing that needs to be done on the data. If we decided to take this approach further, I would recommend using a SQL dump from nzsl-share restored into a SQLite database, or simply adjusting the output from nzsl-dictionary-scripts to write out a database with the full data from the Freelex XML dump (I think that even the nzsl-share cached sign data only includes a minimum of data required). In particular, having IDs and actual picture URLs would reduce the complexity of this feature significantly. We don't have budget to increase the capacity of the Postgres database, so we would still need to store this data in a SQLite file.

**Alternatives:** There are some quick-win alternatives that I've seen around this. Unfortunately, most come at a cost, but it would be unfair of me not to mention them.

1. Static IP - if we had a static IP to give VUW, we would probably not be blacklisted. There are a couple of Heroku addons that provide static IPs - https://elements.heroku.com/addons/quotaguardstatic and https://elements.heroku.com/addons/fixie. Both of these have pretty steep costs for moderate traffic - I think these would run between $50 and $90 USD a month, which basically is the same cost as the entire remaining cost of hosting the applicaiton.
2. Larger Postgres database: Our current database only stores vocab sheets, and has a row limit of 10,000 rows on the free tier. If we bumped up to the hobby tier, we would have enough capacity to cache full sign data just like nzsl-share does. We could use the same rake task used by that application to popular the database. This would cost an additional USD $9/mo, which isn't much, but is still an additional USD $108/mo per year on a hosting budget that is funded in advance two years at a time - e.g. there's very little wiggle room. It's also unclear whether this would gain us anything over a SQLite database performance-wise, since the database is read-only.
3. Feature flag only. This is a quick win in regards to complexity, rather than cost, but we could just have a feature flag to say whether Freelex is enabled or not, and show a banner accordingly, but otherwise not use cached signs. I considered this option, but at the end of the day, without Freelex or some offline signs, the dictionary really isn't useful in any way. All of it's content revolves around sign data, and without Freelex, we can really only show the homepage and some static pages. 
4. A special outage page for Freelex: At the moment, when Freelex isn't available, we throw a 500 error, with the default Rails "Something went wrong" page. We could consider either putting a branded 404/500 page in there (perhaps NZSL Share's [500](https://nzslshare.nz/500.html) or [404](https://nzslshare.nz/404.html) page?), or a page specifically for Freelex outages which could try and explain why the site isn't working (maybe like Cloudflare does, with the diagram showing that the user is connecting to NZSL fine, but NZSL can't connect to the database/dictionary/service/something?). 

More screenshots:

image

image

image


Rabid-Dan (approved):
This seems like a really solid MVP solution to me. Big ups for doing the work on this!

Br3nda:
I'm nervous about the use of sqlite here. The file itself will be gone after a dyno restart (heroku won't keep it), but more worried about it's ability under load.

Br3nda:
But even with my sqlite nerves, what you have here is going to be more reliable than our current Just-Not-Working when freelex is down/blocking-requests.

eoinkelly:
@Br3nda The SQLite file is checked into this repo so it should always be around (unless I'm missing something)?

app/controllers/signs_controller.rb
```diff
@@ -39,6 +41,11 @@ def autocomplete
 
   private
 
+  def find_sign
+    id = permitted_params[:id]
+    freelex_enabled? ? Sign.first(id: id) : OfflineSign.find_by_id!(id) # rubocop:disable Rails/DynamicFindBy
```
eoinkelly, line 46: Could this be `OfflineSign.find_by!(id: id)` which would keep rubocop happy?

Br3nda:
We've hit the slug size limit on heroku  #1135
the sqlite database will add to the size and take us way over.

PR #1098, opened 2020-07-28
Add themed error pages (404, 500, 422) from nzsl-share

joshmcarthur:
Occasionally, we get blacklisted by the VUW network from contacting the backend database. When this happens, visitors are presented with this error page:

image

This is quite jarring, and kind of makes things look a lot more broken than they are (which, to be clear, is still very broken). 

This PR switches in the error pages from NZSL Share, with some adjustments to logos, colours and typography to match NZSL Dictionary. 

404:

image

500:

image

422: 

image

These pages also have nice clear typography if we wanted to add a more detailed error if we knew something was going on - like Freelex being down.

I don't know what happened to the orange in the screenshots above. It is configured to match the orange of the search button.


DSRU:
Is the 422 error relevant to the dictionary? It can only be searched and not edited, so it seems strange to have a message saying 'the changes you requested ... '

joshmcarthur:
@DSRU it's not as _likely_ as in NZSL Share, but there are still forms that can be submitted - for example:

* Adding signs to a vocab sheet
* Removing "
* Any button press that ends up being a POST request is wrapped in a form submission - there are just no form fields

I don't think it's an error page that is likely to be presented very often, but I've definitely seen it happen before, and it seems worthwhile to me to keep the presentation of the error pages all consistent.

Br3nda:
LGTM.

Okay for me to send this to staging @DSRU  ?

Br3nda (approved):

DSRU:
@Br3nda yes, go ahead

eoinkelly (approved):

eoinkelly:
@Br3nda I'll merge and check it goes through ok (I have some time today)

PR #1113, opened 2020-08-19
Upgrade rubocop, rubocop-performance, rubocop-rails

eoinkelly:
Rubocop upgrades always come with new lints and changes to existing lints.
This change upgrades rubocop (and associated gems) and sets values for
the new lints. Default settings are used whenever possible except where
they would require too many changes to the codebase. These exceptions are:

* Rubocop seems to have changed how it calculates AbcSize so we adjust the
  size from 22 to 26 to keep our current code passing
* Style/OptionalBooleanParameter is disabled because the existing code makes
  extensive use of non-optional default params.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -82,7 +82,7 @@ Metrics/MethodLength:
   Max: 20
 
 Metrics/AbcSize:
-  Max: 22
+  Max: 26
 
 Metrics/PerceivedComplexity:
   Max: 12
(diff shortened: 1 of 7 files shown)


.rubocop.yml
```diff
@@ -82,7 +82,7 @@ Metrics/MethodLength:
   Max: 20
 
 Metrics/AbcSize:
-  Max: 22
+  Max: 26
```
Br3nda, line 85: i would be nice to remove this and go down to the default levels

Br3nda:
needs updating in codeclimate.yaml too

Gemfile
```diff
@@ -84,7 +84,7 @@ group :development, :test do
   ##
   # We want to use the same version of rubocop as Codeclimate does - see
   # .codeclimate.yml and https://docs.codeclimate.com/docs/rubocop
-  gem 'rubocop', '~> 0.87.1', require: false
+  gem 'rubocop', '~> 0.89.1', require: false
```
Br3nda: also needs updating in `.codeclimate.yaml`

Can probably go to all the way to 0.90 now

Br3nda (approved):

PR #1130, opened 2020-09-10
Bundle update

Br3nda:
includes rspectre which was used in #1129


Br3nda:
Tests pass, but the code climate key isn't being sent to this test run because I'm on a fork

PR #1137, opened 2020-09-11
Removing percy

Br3nda:
the dev dependencies may or may not be installed on heroku, but
they amount to 800M on dev - so removing to see if it fixes our
too large heroku slug error that is stopping deployment


Br3nda:
passes -

PR #1155, opened 2020-09-17
Production deploy to fix PDF rendering

eoinkelly:
This rolls back Puppeteer to 2.0.0 which has fixed PDF rendering on staging so I'm hopeful will do the same on production.

Puppeteer is never exposed to users and only renders one page which is generated from our application so I'm ok with being on an older version for a while until we can figure out why newer versions don't install properly.


joshmcarthur:
It's a pity JSON doesn't allow comments, it'd be nice to flag to an upgrader that > 2.0 breaks PDF generation in the Heroku dyno

joshmcarthur (approved):

PR #1210, opened 2020-11-18
Bugfix/ccsd 1563 downgrade puppeteer again and also tell dependabot to ignore it

JacobBriggsAckama:
downgrade puppeteer again and also tell dependabot to ignore it


joshmcarthur:
Pretty sure that @eoinkelly and myself looked into the security implications of this and decided it was OK. A significant part of this is to do with _how_ we use Puppetteer, which is to render our own pages under our control as PDF. I'd have more concern if we were rendering third-party content, but we're not.

Br3nda (approved):

PR #1324, opened 2021-04-28
Fix issue causing app to crash if Freelex unavailable.

eoinkelly:
Sign#search was not returning the correct value in the case where an
exception was raised talking to Freelex.


G-Rath (approved):

eoinkelly:
Looks like our test suite fails if the assets server is unavailable. I'm confident in this change helping to mitigate the production outage so I'm going to push it through with those failing specs.

PR #1463, opened 2022-06-02
Use native playback speed control for videos

joshmcarthur:
This pull request updates the support for slow motion videos on sign pages to use the HTML video tag's native playbackRate support, rather than relying on a separate slowed-down video file. This has a number of advantages:

1. The dictionary management software no longer needs to maintain and manually encode slow-motion videos for main gloss videos and examples
2. The storage size of videos will be drastically reduced without the need to store slowed-down videos (not half, because not all videos have slow recordings, but a decent amount do)
3. Users only need to load one video, not two

We also have options in the future to vary the playback speed more dynamically, other than just half-speed or full-speed. I also note that while working on this feature, I noticed that we're really not using VideoJS anymore. It actually doesn't even seem to be properly initialised, but we're still depending on this library. It would be good to clean this up sometime to be clear that developers should expect to work with native video tags rather than the VideoJS API.

Video demo, showing main gloss and usage examples being toggled in and out of slow-motion state, and to show that videos without slow-motion continue to work as normal:

https://user-images.githubusercontent.com/292020/171751100-ba0b0197-5614-4848-af57-f98287e818dd.mp4

### DIFF
--- a/app/assets/javascripts/base.js.erb
+++ b/app/assets/javascripts/base.js.erb
@@ -1,4 +1,4 @@
-$(function(){
+$(function() {
   var setup = function(){
     setup_ckeditor_video_links();
     setup_videos();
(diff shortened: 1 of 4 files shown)


G-Rath (approved):
this is an awesome win!

PR #1465, opened 2022-06-19
ci: setup code scanning workflow

G-Rath:
Actions + Code Scanning is free for public / oss repos, so let's start using that!


JacobBriggsAckama (approved):

G-Rath:
I'm going to ignore the weakened CSRF vulnerabilities that are being flagged for now in order to land this, because they're about changing the order of stuff that's cosmetic so if they're exploited there's not a lot of damage that can be done, and it looks like it'll be a bit of work to fix.

(still would be nice to have them addressed some day)

PR #1480, opened 2022-12-12
fix: update to Ruby 2.7.7 for security

G-Rath:
Does not impact us, but is one less ping from our auditor.


joshmcarthur (approved):
LGTM, as long as the build passes

PR #1481, opened 2022-12-13
NZSL-51: Switch the dictionary application to support using dictionary data from an SQLite file packaged with the app

joshmcarthur:
This pull request builds off of https://github.com/ODNZSL/nzsl-online/pull/1476 to add the Signbank adapter code for the dictionary application. This allows the dictionary to be served either from Freelex, or from a static Signbank database packaged in the repository. We expect these adapters to co-exist until February, at which point the Freelex is decomissioned, and Signbank will continue to be the canonical source for dictionary data.

The scope of this pull request is to have the dictionary support Signbank, preferably in a way that allows the app to revert easily to Freelex in a scenario where we either need to A/B test some missing data (e.g. we might change UAT to run off of Freelex to compare against Signbank data), or in the case we find some key data missing (this shouldn't happen). I feel this PR achieves this scope. What is _not_ in scope is any significant refactoring of existing code. There are plenty of opportunities for refactoring, however the focus here is to add support for Signbank. It's likely that refactoring tasks will become clearer as the Freelex code is eventually removed, making it more obvious where we have call sites that can be tidied by inferring that only data from Signbank needs to be handled, rather than Signbank AND Freelex.

The database structure we [export from Signbank](https://github.com/ODNZSL/nzsl-dictionary-scripts/pull/12) is represented by the following diagram:

NZSL Signbank Export Database

This database is connected to with ActiveRecord models, in the `app/models/signbank` namespace. These models are marked as readonly in every environment except `test` - we don't intend for writes to hit this database, it's just an efficient way to store the dictionary data in a relational way.

These models are consumed, primarily by a new controller for searching and displaying signs, as well as adding to vocab sheets. Functions that used to be implemented within Freelex, such as keyword & filter searches, filtering NSFW signs out of presentational contexts, grabbing a random sign, and autocompleting searches are reimplemented within this codebase. 

In some places, I found dead code, and a couple of bugs, and these have been addressed in this PR:

* The vocab sheet was collapsing it's height to 0px due to float clearing issue. I updated the selector to one that was not affected by the clearfix problem after determining it wasn't practical to just clearfix some more.
* Search images used to switch sizes. They don't any longer, so this code could be removed.
* The pagination service was refactored slightly to not rely on search params to function - it now only needs them for adding to params for next, previous pages etc.

Generally, signbank has been able to co-exist with Freelex. The main place where I found this was difficult or impractical related to locations and location groups. With the existing Freelex code, locations were grouped into 'location groups', and each group had child locations. A location and location group could be inferred by it's number - e.g. 4.10.Neck/Throat would be location 10, in location group 4.  Signbank uses a different reference system, where each location is zero-padded, but does not have a group reference - e.g. the location might be 10 - neck/throat. Because of this, I have to use some historic Freelex data to infer the group by matching against the location number. I also had to duplicate the set of location images so that I can reference them by the name of the location, rather than it's fully qualified identifier, which is why the diff is so full of location imagery changes.

I expect this part of the code to be a good candidate for cleanup once Freelex no longer requires supporting code, where we can switch call sites to just use the Signbank data structures.

Comparative screenshots (note: identical results here are desirable, but the Freelex and Signbank data has diverged, slightly ;-))

#### Homepage 

Production:

www nzsl nz_

Dev:

localhost_3000_

#### Sign page

Production:

www nzsl nz_signs_856

Dev:

localhost_3000_signs_856

#### Search results

Production:

www nzsl nz_signs_search_tag=8

Dev:

localhost_3000_signs_search_tag=House+and+garden

#### Advanced search results

Production:

www nzsl nz_signs_search_s= hs= l= lg=1 tag=8 usage=

Dev:

localhost_3000_signs_search_s= hs= l= lg=1 tag=House+and+garden usage=

#### Vocab sheet

Production:

www nzsl nz_vocab_sheet

Development:

localhost_3000_vocab_sheet

---

There are minor differences in search result counts, and ordering. I'm expecting to introduce smaller patches to resolve these where necessary, but they need more concentrated work between myself and ODNZSL to understand where these differences are coming from, and whether the data from Freelex or our Signbank implementation should be considered correct.


app/assets/javascripts/base.js.erb
```diff
@@ -236,9 +238,6 @@ $(function() {
     var selected_images = container.find('.selected:not(.group.selected .selected)').children('.image');
     selected_images.clone().appendTo(selected_container);
     /* each of the images needs to be pointing to the smaller image. boom. */
-    selected_container.find('img').each(function(){
```
joshmcarthur, line 239: I think this is from an older version of the UI. It causes not found errors anyway, since these images have different asset digest hashes, so just switching the dimension part of the path doesn't do anything. I removed this and tested all search types and didn't notice any problems.

app/assets/javascripts/base.js.erb
```diff
@@ -1,4 +1,6 @@
 $(function() {
+  var SEARCH_DELIMITER = '<%= SearchQuerySanitizationService::DELIMITER %>';
```
joshmcarthur, line 2: We can no longer use a space character as the delimiter, since some fields we search by now contain spaces (e.g. topic tags).

app/assets/javascripts/vocab_bar.js
```diff
@@ -24,8 +24,8 @@ $(document).ready(function() {
       if (barBottom > footerTop) {
         if ($('.search-results__container').length > 0) {
           bar.css({height: $('.search-results__container').innerHeight()});
-        } else if ($('.search-result-container').length > 0) {
```
joshmcarthur, line 27: `.search-result-container` has a wider clearfix issue which results in it having no height. Switched to vocab_sheet_background which a) works and b) seems like a better element to target here.

app/helpers/sign_examples_helper.rb
```diff
@@ -0,0 +1,13 @@
+module SignExamplesHelper
+  def autolink_glosses(sentence, current_sign_id: nil)
+    sanitize(sentence.gsub(/(?<gloss>[^\d\s]+)\[(?<id>\d+)\]/) do
```
joshmcarthur: Replace gloss link notation (`word[1234]`) with a link to the sign page.

app/models/freelex/example.rb
```diff
@@ -0,0 +1,6 @@
+module Freelex
+  class Example
```
joshmcarthur, line 2: This class exists so that we can use the same API across Freelex and Signbank

app/models/item.rb
```diff
@@ -19,14 +19,13 @@ def initialize(attrs = {})
...
 
-    sign = SignModel.resolve.fetch_by_id_via_cache(sign_id)
-
+    sign = SignModel.resolve.find(sign_id)
     return if sign.nil?
 
     self.id         = sign.id          if id.nil?
     self.name       = sign.gloss_main  if name.nil?
     self.maori_name = sign.gloss_maori if maori_name.nil?
-    self.drawing    = sign.drawing     if drawing.nil?
```
joshmcarthur, line 29: While the glosses can be overridden on the vocab sheet, the drawing cannot, so we don't need a condition on this

app/models/sign_model.rb
```diff
@@ -2,7 +2,23 @@
 # This class bridges common operations where the correct class cannot
 # be inferred directly (e.g. not within a model)
 class SignModel
```
joshmcarthur, line 4: We switch SignModel because Signbank has different topic and usage tag values

app/services/signbank/sign_search_service.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+module Signbank
+  class SignSearchService
+    SEARCH_SQL = <<-SQL.squish
+    WITH sign_search(id, rank_precedence, rank_order, row_num) AS
```
joshmcarthur, line 6: This is the same search implementation as NZSL Share, with additional filter searches for handshape, location, usage and topic tags

app/services/signbank/sign_search_service.rb
```diff
@@ -0,0 +1,130 @@
...
+
+    def initialize(query, relation: Sign)
+      @relation = relation
+      @query = query
+    end
+
+    def keyword_search_args
+      term = @query[:s].first
+      unaccented_term = ActiveSupport::Inflector.transliterate(term)
+      { term: unaccented_term, like_term: "% #{Sign.sanitize_sql_like(unaccented_term)} %" }
```
joshmcarthur, line 56: Spaces are wrapped around the term to match whole words. This may need some tuning, but SQLite doesn't have a built-in regexp match function

app/models/signbank/sign.rb
```diff
@@ -0,0 +1,72 @@
...
+    def self.safe_for_work
+      left_outer_joins(:topics)
+        .where("usage IS NULL OR usage != 'obscene'")
+        .group(:id)
+        .having <<~SQL.squish
+          SUM(
+            CASE WHEN word_topics.topic_name IS NOT NULL
+                 AND word_topics.topic_name = 'Sex and sexuality'
+            THEN 1 ELSE 99 END
+          ) % 99 = 0
```
mischa-s, line 26: I mean... I would have thought there'd be a more conventional way to do this but as your SQL is general more advanced than mine, I'm assuming this is the best you were able to come up with?

app/helpers/search_helper.rb
```diff
@@ -0,0 +1,230 @@
...
+        hand_shape.split('.').last == '1',
+        simple
+      )
+    end
+
+    safe_join(selected, ' ')
+  end
+
+  def display_location_groups_search_term(simple = false)
+    locations = ::SignMenu.resolve.location_groups.select do |location_group|
```
mischa-s, line 154: @joshmcarthur is `::` syntactic sugar when used like this?

app/helpers/search_helper.rb
```diff
@@ -0,0 +1,230 @@
+# frozen_string_literal: true
+
+# rubocop:disable  Rails/HelperInstanceVariable - extensive use of @query in these helpers
```
mischa-s, line 3: @joshmcarthur while it's fine... I don't love the use of @query here. Not sure if this is existing code or new code, but would be good if there is a way to make this more boring, especially given how long the file is

mischa-s (commented):
Similarly to the other PR, I don't have much context on this so I am mostly just checking code structure and logical consistence which seems pretty good - have called out a couple of questions I had while reviewing

app/models/signbank/sign.rb
joshmcarthur: @mischa-s SQLite doesn't support the EVERY operator, this was the cleanest way I found to do this. I realise it's not the cleanest, which is why I added so much commentary around it.

app/helpers/search_helper.rb
joshmcarthur: @mischa-s have a read of the description in the PR - refactoring existing behaviour in this app is out of scope. I don't love it either, but it's what is there, and it's not that trivial of a refactor to change all the call sites for these methods to pass in `@query`.

joshmcarthur: And to answer your question - yes, it's existing code.

app/models/signbank/sign.rb
joshmcarthur: Specifically, this fixes an issue that otherwise comes up when a sign has more than one topic, but only one topic is 'Sex and sexuality'. Without this clause, the sign is still included as a candidate, because the LEFT OUTER JOIN (which is necessary to include signs that do not have any topics) includes a result row for each topic the sign has. 

So if a sign has the topics 'Animals' and 'Sex and sexuality', when I tried to do this with a simple WHERE clause, it was still being included in the result set - the row for the sign with the topic 'Sex and sexuality' was excluded, but the row for the sign with the topic 'Animals' was still there, so the sign was still included.

Having said this, I'll try and revisit this since I'm doing some other SQL improvement work today. Perhaps I can find a way to do this with a subselect instead.

app/helpers/search_helper.rb
joshmcarthur: This is leftover from when `SearchHelper` was `Freelex::SearchHelper`. Prefixing a class or module name with `::` causes Ruby to check the top-level scope for the class, instead of the module.

Because we have a class named `Freelex::SignMenu` (which the top-level `SignMenu#resolve` method resolves to when the app is in Freelex dictionary mode), the `::` was necessary when `SearchHelper` was `Freelex::SearchHelper` to ensure that Ruby would find the correct class.

mischa-s (approved):

hopkincame (approved):

PR #1498, opened 2023-02-13
NZSL-115 Removed all old Freelex code

hopkincame:
This PR removes all old code relating to Freelex now that we have fully moved away from that system. This includes updating some of the documentation that referred to Freelex to state that the database is no longer used and moving code from Freelex classes to their subclasses.


app/views/admin/requests/index.html.haml
```diff
@@ -1,4 +1,4 @@
-%h1 Requests to Freelex
+%h1 Requests to Signbank
```
joshmcarthur: We don't need this controller or template at all anymore, since we're not recording any requests.

config/routes.rb
```diff
@@ -27,7 +27,7 @@
 
   root to: 'pages#show'
 
-  resources :signs, only: :show, module: SignModel.adapter.to_sym do
+  resources :signs, only: :show, module: :signbank do
```
joshmcarthur: I'd like to pull these controllers up to the top-level now we don't have multiple adapters

joshmcarthur (changes requested):
There's a bit more consolidation I'd like to do here. I think the models make sense to leave nested in the `Signbank` namespace, since they share a common base model. The controllers I think it would be clearer to move out of the `signbank` module to the top level since they're basically the core of the application.

joshmcarthur (approved):
Nice work! Sorry it took me a few days to come back to this - I wanted to have the time to check out the branch and try breaking things - which I couldn't manage :-)

I pushed one tiny commit - nothing was breaking, but I noticed that when I changed the sitemap generation over, I hadn't removed a Freelex param and a comment that still referred to pulling XML down (which we don't do any more)

PR #1500, opened 2023-02-21
Rails 7 upgrade

hopkincame:
This PR updates rails to version 7 and makes some minor changes to facilitate this upgrade.


Gemfile
```diff
@@ -18,18 +18,12 @@ gem 'jquery-ui-rails'
...
 
-# As of 2020-05-23, the Rails 6 compatible version of ckeditor_rails has not
-# been released to Rubygems so we are instead referencing the commit where the
-# master branch gained Rails 6 support but has the fewest other changes from
-# the released version.
-#
-# It is our intention to revert to the official Rubygems version of the gem as
-# soon as possible.
-gem 'ckeditor_rails', git: 'https://github.com/tsechingho/ckeditor-rails.git', ref: 'dc5de031b0dfde527cf5571bb87b90c59ca24ab1'
+gem 'ckeditor_rails'
```
joshmcarthur, line 26: Just checking that you've accessed the admin and verified that CKEditor is working all good with the new version

Gemfile
```diff
@@ -3,7 +3,7 @@
 source 'https://rubygems.org'
 ruby File.read('.ruby-version')
 
-gem 'rails', '~> 6.1.4'
+gem 'rails', '~> 7.0'
```
G-Rath: This'll allow 7.1+ which we don't want (at least, not to just "happen" as part of someone doing `bundle update rails` for a security patch):

```suggestion
gem 'rails', '~> 7.0.0'
```

G-Rath (changes requested):

Gemfile
joshmcarthur: Nice catch!

Gemfile
hopkincame: It appears to be working 
Screenshot 2023-02-28 110255

G-Rath (approved):

joshmcarthur (approved):

PR #1519, opened 2023-10-03
Remove jquery UI rails gem

hopkincame:
This PR removes the gem jquery-ui-rails due to vulnerabilities and it being unmaintained and replaces it with jquery-ui installed via yarn


G-Rath (approved):
squash when merging

PR #1562, opened 2024-12-09
CC-2240: upgrade to Node v20

G-Rath:
This upgrades us to the latest LTS version of Node, and also switches us to using `.node-version` which is our preference these days


nzlaura (approved):
Looks good! 😄

PR #1567, opened 2025-01-08
CC-2354: upgrade to Ruby 3.2

JacobBriggsAckama:
Upgrade to ruby 3.2, with minimal changes - no bundle updates now, just postgres and sqlite and pry updates needed


.ruby-version
G-Rath, line 1: issue: we should go to the latest patch version, since that includes security fixes

G-Rath (changes requested):
it looks like you've done a `bundle update` which I'm generally not a fan of, but I know you had some issues with `bundler` and CI is passing so I can live with it I guess...

G-Rath (approved):
love it!

PR #1568, opened 2025-01-30
CC-2354: upgrade to Ruby 3.3

JacobBriggsAckama:


G-Rath (changes requested):
looks good, you just need to update the `RUBY VERSION` line in the `Gemfile.lock` too

G-Rath (approved):

PR #1583, opened 2025-04-11
CC-2378: upgrade to rails 7.2

JacobBriggsAckama:
Upgrade to rails 7.2 and supporting gems

Replace secrets with apps.yml

Made the Record class an abstract class, as after the upgrade the tests were failing on "record" not being a table when doing queries - because its not.  Now I don't know what changed between rails 7.0 and rails 7.2 but I went spelunking through active record and in rails 7.0 this model was not being treated as a table that activerecord expected to find in the database, while in rails 7.2 it was being treated as a table it expected to see in the database.

Update database_cleanmer to support rails 7.2, and also specify that we want to use 'database_cleaner-active_record' 

Configure database cleaner to truncate the database before a run, and roll back transactions after each example, as CI was failing due to old data in the database persisting across runs and examples


G-Rath (approved):

config/application.rb
```diff
@@ -12,6 +12,11 @@ module NzslOnline
   class Application < Rails::Application
     config.load_defaults 7.0
```
G-Rath:
```suggestion
    config.load_defaults 7.2
```

G-Rath (changes requested):

G-Rath (approved):

PR #1613, opened 2025-12-16
RRU2026-11: upgrade Rails to v8.1

JacobBriggsAckama:
rails to 8

unpin sqlite3 and puma

eager loading on

fork gem 'ckeditor_rails' to support rails 8 with a one character change - I don't know how to move forward with this but it works as is. I added an `8` to lib/ckeditor-rails.rb here in this fork https://github.com/JacobBriggsAckama/ckeditor-rails/commit/73f79b7876cb8ea7b9478cf8c244f778ee45d1d0

change deprecated on_worker_boot to before_worker_boot

fixture_path to fixture_paths


Gemfile
```diff
@@ -25,15 +25,15 @@ gem 'whenever'
 
 gem 'autoprefixer-rails'
 
-gem 'ckeditor_rails'
+gem 'ckeditor_rails', github: 'JacobBriggsAckama/ckeditor-rails'
```
G-Rath: issue: if we decide to stick with the fork, then we should move it into the Ackama repo

However, it looks like the change is about controlling what [file gets required automatically when the gem is required](https://github.com/JacobBriggsAckama/ckeditor-rails/blob/73f79b7876cb8ea7b9478cf8c244f778ee45d1d0/lib/ckeditor-rails.rb#L87-L94), which feels like something we should be able to do manually?

i.e. add `require 'ckeditor-rails/engine'` to `config/application.rb` or `config/boot.rb`

G-Rath (changes requested):

Gemfile
JacobBriggsAckama: Adding `require 'ckeditor-rails/engine'` to `config/application.rb` allowed me to remove the dependency on my changed repo!

nzlaura (approved):

G-Rath (changes requested):
need to update `config.load_defaults` to 8.1

G-Rath (approved):

PR #1618, opened 2026-02-02
chore: upgrade to eslint v9

nzlaura:
Addresses: https://github.com/advisories/GHSA-p5wg-g6qr-c7cg

Notes: 

I needed to remove `eslint-plugin-google` as it's not compatible with eslint v9 

I've left this using the legacy config rather than upgrading to flat config due to planned work with the NZSL sites.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "doc": "doc"
   },
   "scripts": {
-    "eslint": "./node_modules/.bin/eslint ."
+    "eslint": "ESLINT_USE_FLAT_CONFIG=false ./node_modules/.bin/eslint ."
   },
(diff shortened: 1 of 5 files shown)


JacobBriggsAckama (approved):

package.json
G-Rath, line 1: suggestion: stick the env variable in the `scripts.eslint` script, as we want everyone to be calling that for consistency

(and then once its there, you can remove it from everywhere else)

G-Rath (approved):

package.json
nzlaura: shifted it, didn't realise I could do that initially 🙈

PR #1658, opened 2026-05-21
CCSD-5856: Update footer with new content and point to copyright page for license info

nzlaura:
I've updated the footer with the new content requested to go with the changed license that will be used to replace the creative commons license. 

CleanShot 2026-05-26 at 10 10 59@2x


db/migrate/20250522000000_add_privacy_page_and_update_copyright.rb
```diff
@@ -0,0 +1,19 @@
...
+        template: 'standard',
+        show_in_nav: true
+      )
+    end
+
+    copyright_page = Page.find_by(slug: 'copyright')
+    copyright_page&.update!(title: 'Copyright and license conditions - VUW NZSL Resources')
+  end
+
+  def down; end
```
JacobBriggsAckama: Why not remove the on the way down?

db/migrate/20250522000000_add_privacy_page_and_update_copyright.rb
G-Rath, line 1: issue: data loading is usually done via rake task or seed, rather than migrations

~[it looks like there is even something built around seeding pages](https://github.com/ODNZSL/nzsl-online/blob/main/db/seeds/pages.csv), though I'm not sure if it's safe to re-run...~ sounds like you've already tried that so I guess... umm.. try again? 😅

G-Rath (changes requested):

db/migrate/20250522000000_add_privacy_page_and_update_copyright.rb
nzlaura: the content differs between the seeds + the content on staging and production. so my assumption is that they're not safe to re run. 

I will double check locally, because that was my initial instinct, to update the `page_parts` then re run `db:seeds`

G-Rath (approved):

PR #1659, opened 2026-05-26
chore: update seeds to only run in test and update content  

nzlaura:
Given that admins are able to edit content in the admin panel, having seeds that overwrite the content by pulling info from `page_parts.csv` and `pages.csv` isn't something we want to run in productio, because it's not able to be re run safely. 

I've also updated the `page_parts.csv` and `pages.csv` to include the content I've added for local development


nzlaura:
@G-Rath should be ready for review now 😄

db/seeds.rb
```diff
@@ -2,4 +2,4 @@
 
 require 'csv'
 
-SeedDataService.load_all
+SeedDataService.load_all if Rails.env.test?
```
G-Rath: this way your dev env will get seeded with content out of the box too

```suggestion
SeedDataService.load_all if Rails.env.local?
```

G-Rath (approved):

db/seeds.rb
nzlaura: I wondered if it was possible in this codebase because it was causing a rubocop failure, but it looks like it's because of the version of rubocop we're on 

It might be worth upgrading rubocop at some point, but I feel like that's a separate task so I've just ignored the cop for the moment.

PR #1662, opened 2026-06-15
staging -> production

JacobBriggsAckama:


nzlaura:
Another one I need to double check with Micky on, @JacobBriggsAckama 😅 I think the NZSL share + online production releases will almost be a pair with my work. Apologies for blocking prod deploys 🙃

G-Rath (approved):

nzlaura (approved):

PR #1665, opened 2026-06-18
CCSD-5856: update app and production config to switch to mailgun

nzlaura:
Emails are not currently sending from the feedback form. NZSL Share and Signbank are using Mailgun, so it makes sense to switch NZSL Dictionary to use mailgun as well rather than sendgrid. 

This PR makes the changes to the app and production config to use mailgun for email sending. It also adds an env var for `CONTACT_EMAIL` so we can update the contact email if needed.


nzlaura:
I think this should be what's needed, so it would be a good to get a review in case there's something I've missed.

JacobBriggsAckama (approved):

PR #1673, opened 2026-07-06
feat: supporting using `turnstile` to reduce spam in feedback form

nzlaura:
This PR adds in turnstile as a captcha equivalent to reduce the amount of spam being sent through feedback forms. 

The fallback keys are the official cloudflare test keys, as `rspec` kept failing in ci when there were no keys present (this wasn't happening locally).

### DIFF
--- a/config/initializers/cloudflare_turnstile.rb
+++ b/config/initializers/cloudflare_turnstile.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+CloudflareTurnstile.configure do |config|
+  config.site_key = ENV["TURNSTILE_SITE_KEY"]
+  config.secret_key = ENV["TURNSTILE_SECRET_KEY"]
+end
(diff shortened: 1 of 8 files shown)


config/initializers/cloudflare_turnstile.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+RailsCloudflareTurnstile.configure do |config|
```
G-Rath, line 3: suggestion: can you add a comment w/ a link to the [page about these](https://developers.cloudflare.com/turnstile/troubleshooting/testing/)

G-Rath (approved):
approved for when you add the comment

G-Rath:
actually after thinking about it, can we use the "always fails" test key? since this is only enabled in production, that should catch if we ever mistakenly have it enabled without configuring proper keys

nzlaura:
@G-Rath I will add the comment and the always fails test key- that makes sense. It was failing in ci only because of the lack of a key, so should be fine 😄

PR #1675, opened 2026-07-06
fix: form returning 500 errors with turnstile enabled

nzlaura:
Initially I tried to implement turnstile without using a gem to help, and when I switched to using a gem,  I missed updating the controller to use the helpers from the library, resulting in 500 errors when you try to submit feedback via the form. 

This change uses the helpers from the gem I used to help implement turnstile for the form: https://github.com/instrumentl/rails-cloudflare-turnstile/blob/main/lib/rails_cloudflare_turnstile/controller_helpers.rb

I updated to using the "always pass" keys from cloudflare to test that this works locally 

CleanShot 2026-07-07 at 08 26 09@2x

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class FeedbackController < ApplicationController
+  include RailsCloudflareTurnstile::ControllerHelpers
+
   before_action :find_vocab_sheet, :set_search_query, :footer_content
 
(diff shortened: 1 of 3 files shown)


G-Rath (approved):

nzlaura:
I feel like I made enough changes here to merit asking for a re review 😅 getting tests to pass when I had to push up to CI to check if my fix worked 🙃

G-Rath (approved):

PR #1698, opened 2026-08-25
CC-2722: update puppeteer for security

G-Rath:
The only breaking change that should impact us is that Puppeteer v19 switched to using a global cache directory which isn't included in the Heroku slug so causes runtime errors; we need to manually copy the cache into the app path so it is captured properly (see https://github.com/puppeteer/puppeteer/pull/9095 for more).

I have also updated the `render-pdf` script with modern changes we've made over the years that should make the rendering more stable.

Addresses GHSA-jmr9-qjv8-65gv


nzlaura (approved):
Looks good

PR #1704, opened 2026-08-25
CC-2722: downgrade puppeteer

G-Rath:
For some reason this is broken on staging, and I can't figure out why so I'm going to revert it to unblock doing some house tidying (like switching to NPM and updating Rubocop) before then trying to land the upgrade in jumps to hopefully pin down the version that's the issue

This reverts commit a1d2580457b5d168102479a0d81c701a0b06ded4.


G-Rath:
Not needed now thanks to #1705

PR #1720, opened 2026-09-15
CC-1157: Replace paperclip

hopkincame:
This PR replaces the deprecated Paperclip with ActiveStorage. As Paperclip was only being used for feedback, the amount of code that needed replacing was actually quite small. I have also added specs to check that videos are added successfully, which required moving "MAIL_FROM" in the config to default instead of just having it in production.

This PR doesn't include dropping the old paperclip columns, I want to see this working in staging etc before doing that.


Gemfile
```diff
@@ -111,6 +107,7 @@ group :test do
   gem "capybara"
   gem "capybara-screenshot"
   gem "capybara-selenium"
+  gem "climate_control"
```
hopkincame, line 110: The PDF Rendering spec uses this gem. This gem was previously included as a dependency of Paperclip. Because of this, removing paperclip also removed this gem. I have now explicitly added this here

db/migrate/20260914043851_migrate_feedback_videos_to_active_storage.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
```
hopkincame: This might not be needed given videos are attached to emails and then deleted, but this wouldn't hurt to have and it means that in the extremely unlikely case of this being deployed as someone is leaving feedback, things won't break

G-Rath: Seems like a good idea to have around at least while we're deploying it, but we should do it as a rake task rather than a migration

G-Rath (changes requested):
awesome how little work this required! I think it's good to go, after you move the data stuff into a raketask

